PR #1, opened 2019-11-10
Create v0.0.1

dfed:
This PR spins up this Package with a working public API and tests.

The `CacheAdvance` is a disk-persistent cache that enables performant serialization of individual messages to disk. It is allocated with a specific byte size, and can either notify the caller when it is full, or overwrite old logs with new ones.

The `CacheAdvance` assumes its methods are called from the same serial thread – it does not handle threading and locking. (I'm open to feedback on this one.)

Requirements for this cache were:
- allows for incremental appends to disk
- has a maximum size
- can write new messages over older messages when it is out of space
- can handle arbitrary structured data
- can fetch the cached data

A PR to update the README will come soon 😬

This code was heavily inspired by Aardvark's [NSFileHandle additions](https://github.com/square/Aardvark/blob/master/CoreAardvark/Private%20Categories/NSFileHandle%2BARKAdditions.m) and [ARKDataArchive](https://github.com/square/Aardvark/blob/master/CoreAardvark/ARKDataArchive.m) written by @pwesten.

Curious what you think @bachand, @fdiaz, @jqsilver, @martinmroz


.gitignore
```diff
@@ -1,12 +1,7 @@
 # Xcode
```
dfed: This file was copied over from [Relativity](https://github.com/dfed/Relativity/blob/master/.gitignore)

Contributing.md
```diff
@@ -0,0 +1,15 @@
+### One issue or bug per Pull Request
```
dfed: This file was stolen from [Relativity](https://github.com/dfed/Relativity/blob/master/Contributing.md). I'm aware I broke my own rule on small PRs 😬

Scripts/ci.sh
```diff
@@ -0,0 +1,25 @@
...
+if [ $ACTION == "swift-package" ]; then
+  swift package generate-xcodeproj --output generated/
+  if [ -n "$DESTINATION" ]; then
+    xcodebuild -project generated/CacheAdvance.xcodeproj -scheme "CacheAdvance-Package" -sdk $SDK -destination "$DESTINATION" -configuration Release -PBXBuildsContinueAfterErrors=0 build
+  else
+    xcodebuild -project generated/CacheAdvance.xcodeproj -scheme "CacheAdvance-Package" -sdk $SDK -configuration Release -PBXBuildsContinueAfterErrors=0 build
+  fi
+fi
+
+if [ $ACTION == "pod-lint" ]; then
```
dfed: This action isn't in use yet.

Scripts/ci.sh
```diff
@@ -0,0 +1,25 @@
...
+  swift package generate-xcodeproj --output generated/
+  pushd generated/
+  bundle exec pod lib lint --verbose --fail-fast --swift-version=$SWIFT_VERSION
+  popd
+fi
+
+if [ $ACTION == "carthage" ]; then
```
dfed: This action isn't in use yet

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// A performant on-disk cache that can be appended to one element at a time.
```
martinmroz: A performant on-disk single-producer, single-consumer log store supporting single-element mutation.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+public final class CacheAdvance<T: Codable> {
+
+    // MARK: Initialization
+
+    /// Creates a CacheAdvance.
```
martinmroz: Creates a new instance of the receiver.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+    /// - Parameters:
+    ///   - file: The URL for the on-disk cache to live.
```
martinmroz: The file URL indicating the desired location of the on-disk store.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+    ///   - maximumBytes: The maximum size of the cache, in bytes.
```
martinmroz: What happens when this size is exceeded and rolling is disabled?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+    ///   - shouldRoll: When `true`, new logs will overwrite the oldest logs when the cache runs out of space.
```
martinmroz: Once the on-disk store exceeds `maximumBytes`, new entries will replace the oldest entry.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+    public init(
+        file: URL,
+        maximumBytes: UInt32,
+        shouldRoll: Bool)
+        throws
+    {
+        self.file = file
+        self.maximumBytes = maximumBytes
```
martinmroz: Without persisting these fields in the file itself, you risk a bad migration where the on-disk store exceeds the maximum file size and rolling is enabled. Not sure what that would mean, maybe nothing -- it may just mean that you end up losing a number of log events at once.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+        self.shouldRoll = shouldRoll
+
+        writer = try FileHandle(forWritingTo: file)
+        reader = try FileHandle(forReadingFrom: file)
+
+        lengthOfMessageSuffix = {
+            if shouldRoll {
+                /// We store both the `endOfNewestMessageMarker` and `offsetInFileOfOldestMessage` after each message.
```
martinmroz: I personally avoid `we` and use write in the other tense whose name escapes me right now haha. "The message suffix requires space for both `endOfNewestMessageMarker` and `offsetInFileOfOldestMessage` after each message when rolling is enabled"

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+                return 2 * Data.messageSpanLength
+            } else {
+                /// We store a `endOfNewestMessageMarker` after each message.
+                return Data.messageSpanLength
+            }
+        }()
+    }
+
+    deinit {
+        try? writer.close()
```
martinmroz: Is it worth closing the reader?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
...
+            return true
+
+        } else {
+            // We're out of room.
+            return false
+        }
+    }
+
+    /// Fetches all messages from the cache.
+    public func cachedMessages() throws -> [T] {
```
martinmroz: IMO returning an Iterable type makes more sense here, since the store could be enormous. Feels like it could be too expensive / unnecessary to actually read back everything at once. Not sure what the state of this is in Swift these days but `-> CacheIterator` where `CacheIterator : IteratorProtocol` that walks over all cached entries would be nice, and if you want them all, you could just `collect()` them into an `[T]`.

Sources/CacheAdvance/DataExtensions.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    /// The width of the encoded message size. If this value is changed, any previously persisted message encodings will not be readible.
+    static let messageSpanLength = MemoryLayout<UInt32>.size
+
+    /// A marker written at the end of the newest message written to disk.
+    static let endOfNewestMessageMarker = Data(UInt32.max)
+
+    /// Initializes Data from a UInt32. The data will always be of length Data.messageSpanLength.
+    /// - Parameter value: the unsigned 32-bit integer to encode as data.
+    init(_ value: UInt32) {
```
martinmroz: IMO you should just use a UInt64, it's big enough for everything™

Sources/CacheAdvance/DataExtensions.swift
```diff
@@ -0,0 +1,34 @@
...
+        var valueToEncode = value
```
martinmroz: This should probably be done with consistent endianness for maximum compatibility. Traditionally, you'd use network byte order (big endian) and wrap this in both directions in `init(bigEndian: UInt64)` and `.bigEndian`. Or little, but either way.

Sources/CacheAdvance/EncodableMessage.swift
```diff
@@ -0,0 +1,52 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// A struct that encodes a message of type T into a data that can be incrementally persisted to disk.
```
martinmroz: `incrementally` may not be the right choice of word.

Sources/CacheAdvance/UInt32Extensions.swift
```diff
@@ -0,0 +1,35 @@
...
+    /// - Parameter data: A data blob representing a UInt32. Must be of length `Data.messageSpanLength`.
+    init?(_ data: Data) {
+        guard data.count == Data.messageSpanLength else {
+            // Data is of the incorrect size and can't represent a UInt32.
+            return nil
+        }
+        let decodedSize = withUnsafePointer(to: data) {
+            return UnsafeRawBufferPointer(start: $0, count: MemoryLayout<UInt32>.size)
+        }
+        self = decodedSize.load(as: UInt32.self)
```
martinmroz: I'd again handle the endianness conversion here and switch to 64-bit integers.

.gitignore
```diff
@@ -1,12 +1,7 @@
...
-#
-# gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
-
-## Build generated
+.DS_Store
```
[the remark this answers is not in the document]
dfed: I haven't noticed an issue with it before. But your question made me look it up 😂

According to [git's documentation](https://git-scm.com/docs/gitignore):
> "`**/foo`" matches file or directory "`foo`" anywhere, the same as pattern "`foo`".

Seems like they're equivalent

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,194 @@
+//
+//  Created by Dan Federman on 11/9/19.
+//  Copyright © 2019 Dan Federman.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
dfed: Dual-licensing seems legally hairy. I'd be happy to fully move to `MIT`, though I think I'll do that in a separate PR.

Sources/CacheAdvance/CacheAdvance.swift
dfed: 😂 brings me back to the objc days

Sources/CacheAdvance/CacheAdvance.swift
dfed: I'll document that maximumBytes should not change for a file that is already persisted.

.gitignore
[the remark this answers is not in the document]
martinmroz: :the-more-you-know:

Sources/CacheAdvance/CacheAdvance.swift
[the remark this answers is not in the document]
martinmroz: Up to you! I've had requests for that on some of my open source crates.

Sources/CacheAdvance/CacheAdvance.swift
martinmroz: Makes sense. You could also just make an interface that consumes a single item, and have another one that consumes all of them via the new interface.

Sources/CacheAdvance/DataExtensions.swift
martinmroz: Yep, that's right.

Sources/CacheAdvance/CacheAdvance.swift
dfed: badness. I've added a clarifying comment (and made the API on appending nicer)

Sources/CacheAdvance/CacheAdvance.swift
dfed: Indeed! These improvements are in the We’ll Get There™ camp 🙂

Sources/CacheAdvance/DataExtensions.swift
dfed: Sweet. I've added, though if this ever happens I'll be kinda shocked 😂

dfed:
This PR is ready for review (i.e. I've stopped pushing to it every few minutes 😂)

Sources/CacheAdvance/CacheAdvance.swift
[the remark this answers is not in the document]
bachand: I feel like it's probably good to stick to one license.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,191 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// A cache that enables the performant persistence of individual messages to disk.
+/// This cache is intended to be written from and appended to from the same serial queue.
```
bachand: "written from and appended to from" -> "written and appended to from"?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,191 @@
...
+        self.file = file
+        self.maximumBytes = maximumBytes
+        self.shouldOverwriteOldMessages = shouldOverwriteOldMessages
+
+        writer = try FileHandle(forWritingTo: file)
+        reader = try FileHandle(forReadingFrom: file)
+
+        lengthOfMessageSuffix = {
+            if shouldOverwriteOldMessages {
+                /// The message suffix requires space for both `endOfNewestMessageMarker` and `offsetInFileOfOldestMessage` after each message when rolling is enabled.
```
bachand: `//`?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,191 @@
...
+                return Bytes(Data.messageSpanLength + Data.oldestMessageOffsetLength)
+            } else {
+                /// The message suffix requires space for `endOfNewestMessageMarker` after each message.
```
bachand: `//`?

Sources/CacheAdvance/Bytes.swift
```diff
@@ -0,0 +1,49 @@
...
+        let decodedSize = withUnsafePointer(to: data) {
+            return UnsafeRawBufferPointer(start: $0, count: MemoryLayout<Bytes>.size)
+        }
+        self = NSSwapBigBytesToHost(decodedSize.load(as: Bytes.self))
+    }
+
+    /// Converts megabytes into the receiver. Will never overflow.
+    /// - Parameter megabytes: The number of megabytes to convert.
+    init(megabytes: UInt8) {
+        self = Bytes(megabytes) * 1000000
```
bachand: I always get this confused by should this be 1024^2 or 1000^?
Some discussion of that choice in a comment could be nice.

Sources/CacheAdvance/CacheAdvance.swift
bachand: It's unclear how `endOfNewestMessageMarker` and `offsetInFileOfOldestMessage` relate to `Bytes(Data.messageSpanLength + Data.oldestMessageOffsetLength)`

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,191 @@
...
+    public let file: URL
+
+    /// Appends a message to the cache.
+    /// - Parameter message: A message to write to disk.
+    public func append(message: T) throws {
+        try setUpFileHandlesIfNecessary()
+
+        let encodedMessage = EncodableMessage(message: message, encoder: encoder)
+        let messageData = try encodedMessage.encodedData()
+        let cacheHasSpaceForNewMessage = writer.offsetInFile + Bytes(messageData.count) + lengthOfMessageSuffix <= maximumBytes
```
bachand: What do you think about making this two lines or making it a property or method?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,191 @@
...
+            try writer.seek(toOffset: 0)
+
+            // We know the oldest message is at the beginning of the file, since we just tossed out the rest of the file.
+            try reader.seek(toOffset: 0)
+
+            // We know we're about to overwrite the oldest message, so advance the reader to the second oldest message.
+            try reader.seekToNextMessage(shouldSeekToOldestMessageIfFound: false, cacheOverwritesOldMessages: true)
+
+            // Start writing from the beginning of the file.
+            try write(messageData: messageData)
```
bachand: I wonder if all of this should be a method?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,191 @@
...
+        while shouldOverwriteOldMessages
+            && writer.offsetInFile < reader.offsetInFile
+            && reader.offsetInFile < writer.offsetInFile + messageLength
+        {
+            // We are a rolling cache. The current position of the writer is before the oldest message,
+            // and writing this message would write into the current message. Advance to the next message.
+            try reader.seekToNextMessage(shouldSeekToOldestMessageIfFound: true, cacheOverwritesOldMessages: shouldOverwriteOldMessages)
+        }
+
+        try writer.__write(messageData, error: ())
```
bachand: What's going on with passing `()` to `error:`?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,191 @@
...
+            try write(messageData: messageData)
+
+        } else {
+            // We're out of room.
+            throw CacheAdvanceWriteError.messageDataTooLarge
+        }
+    }
+
+    /// Fetches all messages from the cache.
+    public func cachedMessages() throws -> [T] {
```
bachand: Maybe just `messages`? Not sure "cached" adds much value.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,191 @@
...
+    /// - `messageData` is an `EncodableMessage`'s encoded data.
+    /// - `endOfNewestMessageMarker` is a big-endian encoded `MessageSpan` of length `messageSpanLength`.
+    /// - `offsetInFileOfOldestMessage` is a big-endian encoded `Bytes` of length `oldestMessageOffsetLength`.
+    ///
+    /// By the time this method returns, the `writer`'s `offsetInFile` is always set to the
+    /// beginning of the written `endOfNewestMessageMarker`, such that the next message
+    /// written will overwrite the marker.
+    ///
+    /// - Parameter messageData: an `EncodableMessage`'s encoded data. Must be smaller than both `maximumBytes` and `MessageSpan.max`.
+    private func write(messageData: Data) throws {
```
bachand: Right now both `append(message:)` and `write(messageData:)` have different behavior based on the value `shouldOverwriteOldMessages`. I wonder if there would be a way to refactor to avoid checking `shouldOverwriteOldMessages` in this lower-level method. IMHO checking `shouldOverwriteOldMessages` in both places makes it a bit harder to reason about since it's harder to draw to think about the code linearly (you need to keep conditionals in your head).

bachand: Maybe it it's as simple as breaking this method apart into multiple.

Sources/CacheAdvance/CacheAdvance.swift
bachand: Also where do these underscores come from?

Sources/CacheAdvance/DataExtensions.swift
```diff
@@ -0,0 +1,42 @@
...
+
+import Foundation
+
+extension Data {
+
+    /// The width of the encoded message size.
+    static let messageSpanLength = MemoryLayout<MessageSpan>.size
+
+    /// A marker written at the end of the newest message written to disk.
+    static let endOfNewestMessageMarker = Data(MessageSpan.max)
```
bachand: I feel like this could be one of the few places in code where it's good to pick a magic number. I think picking a distinctive number that isn't something you ever see can be valuable in cases like this. It reduces the likelihood of writing it by accident.

Speaking of that... is there any risk that someone could have a message payload that would contain just the wrong type of data that would cause us to get confused about where the end is?

If that's a possibility I wonder if there would be value in storing your metadata in a header section in the file?

Sources/CacheAdvance/CacheAdvance.swift
dfed: Changing to: `This cache is intended to be written to and read from the same serial queue.` "written from" and "appended to" are the same freaking action 🤦‍♂️. Thanks for calling this out.

Sources/CacheAdvance/CacheAdvance.swift
dfed: I've started using `///` in inline comments, since it makes for better formatting of code within back-tics.

dfed: Re the size correlation, that's a good call-out. Will try to make that more clear.

Sources/CacheAdvance/Bytes.swift
dfed: 😅 I def did some wikipedia digging on this one before choosing this number. I'll add a comment.

Sources/CacheAdvance/CacheAdvance.swift
dfed: Good questions! These all come from Apple's API:
image

It looks like the original, non-`__` versions all threw objective-c exceptions, not Swift errors, and are now unofficially deprecated:
image

It is indeed super-weird that `__write` takes a `Void` error... but I'm just following what the API declares.

Sources/CacheAdvance/DataExtensions.swift
dfed: `MessageSpan.max` _is_ a magic number 🙂

> I think picking a distinctive number that isn't something you ever see can be valuable in cases like this. It reduces the likelihood of writing it by accident.

I like how you're thinking about this. The good news is I've already prevented writing it accidentally. Before you append a message, I have the following check: `messageLength <= Bytes(MessageSpan.max)`. Because I always write the message length span before the message itself, it's impossible to to accidentally write `endOfNewestMessageMarker` into the space where we write the message length span.

By picking `MessageSpan.max` as the magic value, we can clearly document what the maximum length of a message is. (It occurs to me that I could potentially use `Data(MessageSpan.min)` -- i.e. `0` -- as the delimiter. I'll investigate that separately.)

That said, I think I need to make it more clear in the code via comments how we're enforcing that we don't accidentally write the value.

> If that's a possibility I wonder if there would be value in storing your metadata in a header section in the file?

Good question! I considered doing this. In the end, it seemed simpler to write this value inline. Otherwise I'd need a separate reading/writing file handle read/write these values at the beginning (or end?) of the cache file.

dfed: Looks like `Data(MessageSpan.zero)` works as the delimiter. I guess the question now is:

Which message is more useful to support:
1. An empty message
2. A message of extreme length

I'm leaning towards #2. Empty log messages aren't terribly useful. I think?

Sources/CacheAdvance/CacheAdvance.swift
dfed: I'll break `Bytes(messageData.count) + lengthOfMessageSuffix` into it's own method, since I use that in multiple spots.

dfed:
> I think it's good that this cache has two options (overwrite or not) but it makes the code harder to follow. I wonder if there is a way that we can refactor in order to diverge the two different code paths more, obviously sharing common logic between them. It would be nice to be able to more easily draw a line through each code path.

Agreed. What I'm grappling with is the amount of overlap between the functionality required to implement two options. I'll play with splitting these things apart.

> I don't fully understand the format yet but at a high level it seems like you are writing the cache metadata inline, such as using a special last message token. I said this in another comment, but is there any risk that someone could construct a message payload that confuses this cache into not thinking the last message is in a different place? For example, if someone had MessageSpan.max in their message, could that cause confusion? I'm wondering if we want a header section (i.e. N bytes at the top of the file that contain metadata). I know this is how some other file formats work.
Is there a way to see code coverage with the tests on github.com?

As discussed in the other thread, the file's design is predictable enough that it is impossible for the metadata to be accidentally written _in a place where it could cause confusion_. Because we always start parsing this file from index 0, and always advance one message at a time, a marker's value can be written as part of a message's encoding without it ever confusing the cache.

> Is there a way to see code coverage with the tests on github.com?

Not that I'm aware of. Maybe @qing knows?

That said, here are my current code coverage numbers:
image
I'll try to up them.

Sources/CacheAdvance/CacheAdvance.swift
dfed: I was able to pull the common code-paths out earlier into the `write(message: T)` method and make this method more targeted.

Sources/CacheAdvance/DataExtensions.swift
dfed: Good news: `try encoder.encode("")` returns an encoded blob of length > 0. Looks like an empty message would be a weird form of data corruption.

dfed:
New coverage numbers:
image

All the untested code are error cases that I don't have any clue how to hit. `Bytes` is low because I don't have a test case for megabytes, since that math is pretty simple.

Sources/CacheAdvance/Bytes.swift
```diff
@@ -0,0 +1,49 @@
...
+/// A storage unit that counts bytes.
+/// - Warning: If this value is changed, previously persisted message encodings will not be readible.
+public typealias Bytes = UInt64
+
+extension Bytes {
+
+    /// Initializes the receiver from a data blob.
+    /// - Parameter data: A data blob representing Bytes. Must be of length `Data.oldestMessageOffsetLength`.
+    init?(_ data: Data) {
+        guard data.count == Data.oldestMessageOffsetLength else {
```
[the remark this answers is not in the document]
dfed: I've renamed it to `bytesStorageLength`.

dfed:
@bachand @martinmroz thank you both for your thorough reviews!

I've addressed both of your feedback. I'd love another review (and approval, if the code warrants it).

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+        // Create data to write from combined message and suffix.
+        var dataToWrite = messageData + Data.endOfNewestMessageMarker
+        if let offsetInFileOfOldestMessage = offsetInFileOfOldestMessage, offsetInFileOfOldestMessage.count > 0 {
+            dataToWrite += offsetInFileOfOldestMessage
+        }
+
+        // Calculate where the message ends so we can seek back to it later.
+        let endOfMessageOffset = writer.offsetInFile + Bytes(messageData.count)
+
+        // Write the complete messge data atomically.
```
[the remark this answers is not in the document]
bachand: When we spoke in person, we weren't yet writing things atomically. Are we doing that now?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+    /// `[messageData][endOfNewestMessageMarker][offsetInFileOfOldestMessage]`
+    ///
+    /// - `messageData` is an `EncodableMessage`'s encoded data.
+    /// - `endOfNewestMessageMarker` is a big-endian encoded `MessageSpan` of length `messageSpanStorageLength`.
+    /// - `offsetInFileOfOldestMessage` is a big-endian encoded `Bytes` of length `bytesStorageLength`.
+    ///
+    /// By the time this method returns, the `writer`'s `offsetInFile` is always set to the beginning of the written `endOfNewestMessageMarker`, such that the next message written will overwrite the marker.
+    ///
+    /// - Parameters:
+    ///   - messageData: an `EncodableMessage`'s encoded data. Must be smaller than both `maximumBytes` and `MessageSpan.max`.
```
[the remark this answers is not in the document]
bachand: On a similar note I wonder if this method should be asserting if it's being asked to write beyond the end of the file.

Sources/CacheAdvance/CacheAdvanceErrors.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+public enum CacheAdvanceReadError: Error, Equatable {
+    /// Thrown when the cache file is of an unexpected format.
+    /// A corrupted file should be deleted. Corruption can occur if an application crashes while writing to the file.
```
bachand: > Corruption can occur if an application crashes while writing to the file.

From a comment in `CacheAdvance.swift` it seems like we are writing atomically. Is this true still?

Sources/CacheAdvance/DataExtensions.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+extension Data {
```
bachand: As I mentioned in an earlier comment, I wonder if there is a more semantic namespace for some of these comments.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+    /// - Parameters:
+    ///   - file: The file URL indicating the desired location of the on-disk store. This file should already exist.
+    ///   - maximumBytes: The maximum size of the cache, in bytes. Logs larger than this size will fail to append to the store.
+    ///   - shouldOverwriteOldMessages: When `true`, once the on-disk store exceeds maximumBytes, new entries will replace the oldest entry.
+    ///
+    /// - Warning: `maximumBytes` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache leads to undefined behavior.
+    public init(
+        file: URL,
+        maximumBytes: Bytes,
+        shouldOverwriteOldMessages: Bool)
```
bachand: I believe we spoke about this in person, but I can't remember what the answer was...
Let's say I create a `CacheAdvance` file with `shouldOverwriteOldMessages == true` and then I open it later with `shouldOverwriteOldMessages == false` (or vice versa).
How will we fail, if it all?
Is there a chance that we _think_ the data is valid and then read garbage data?

Sources/CacheAdvance/EncodableMessage.swift
```diff
@@ -0,0 +1,52 @@
...
+/// - `data` is length `messageSize`.
+struct EncodableMessage<T: Codable> {
+
+    // MARK: Initialization
+
+    /// Initializes an encoded message from the raw, codable source.
+    /// - Parameters:
+    ///   - message: The messages to encode.
+    ///   - encoder: The encoder to use.
+    init(message: T, encoder: JSONEncoder) {
```
bachand: Random nerd thought: I suspect that writing this as a binary property list is more space efficient.

Sources/CacheAdvance/EncodableMessage.swift
```diff
@@ -0,0 +1,52 @@
...
+        self.encoder = encoder
+    }
+
+    // MARK: Internal
+
+    /// The encoded message, prefixed with the size of the message blob.
+    func encodedData() throws -> Data {
+        let messageData = try encoder.encode(message)
+        let encodedSize = Data(MessageSpan(messageData.count))
+        return encodedSize + messageData
```
bachand: Wow I didn't know you can add `Data` like this. So this just concatenates the data?

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -0,0 +1,160 @@
...
+            // The oldest message is always at the beginning of the cache.
+            try seek(toOffset: 0)
+        }
+    }
+
+    /// Seeks to the next message. Returns `true` when the span skipped represented a message.
+    /// When `false` is returned, it signifies that the last message marker was passed.
+    ///
+    /// - Parameters:
+    ///   - shouldSeekToOldestMessageIfFound: When `true`, the file handle will seek to the oldest message if the last message marker is pased.
```
bachand: "passed"

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -0,0 +1,160 @@
...
+            } else {
+                // Seek back to where we started.
+                try seek(toOffset: startingOffset)
+            }
+            return false
+
+        case .emptyRead:
+            // The remaining file is empty.
+            if shouldSeekToOldestMessageIfFound {
+                // The oldest message is at the beginning of the file.
```
bachand: Do we only get back an `.emptyRead` if we have a cache that will never overwrite old messages?
If we can get back `.emptyRead` for a cache that _does_ overwrite old message, it seems like we can't assume the oldest message is at the start of the file.

bachand: If this behavior is expected, maybe there is a way that we can clarify.

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -0,0 +1,160 @@
...
+                try seek(toOffset: 0)
+            } else {
+                // Seek back to where we started.
+                try seek(toOffset: startingOffset)
+            }
+            return false
+
+        case let .span(messageLength):
+            // There's a valid message here. Seek ahead of it.
+            try seek(toOffset: offsetInFile + UInt64(messageLength))
```
bachand: Should we be using one of our type aliases (e.g. `Bytes`) instead of `UInt64`?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+
+    /// Advances the reader until there is room to store a new message without writing past the reader.
+    /// - Parameter messageLength: the length of the next message that will be written.
+    private func prepareReaderForWriting(dataOfLength messageLength: Bytes) throws {
+        while writer.offsetInFile < reader.offsetInFile
+            && reader.offsetInFile < writer.offsetInFile + messageLength
+        {
+            // The current position of the writer is before the oldest message, which means that writing this message would write into the current message.
+            // Advance to the next message.
+            try reader.seekToNextMessage(shouldSeekToOldestMessageIfFound: true, cacheOverwritesOldMessages: true)
```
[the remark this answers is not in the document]
bachand: It makes more sense to me after reading the code of that method, but it took awhile to grok.

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -0,0 +1,160 @@
...
+
+        return .span(messageSize)
+    }
+
+}
+
+private enum NextMessageSpan {
+    case span(MessageSpan)
+    case emptyRead
+    case endOfNewestMessageMarker(offsetOfFirstMessage: Bytes)
```
bachand: I wonder if we should say `offsetOfOldestMessage` to be more symmetric?

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -0,0 +1,160 @@
...
+            return true
+
+        case .invalidFormat:
+            throw CacheAdvanceReadError.fileCorrupted
+        }
+    }
+
+    // MARK: Private
+
+    /// Returns the next encoded message span, seeking to the end the span.
```
bachand: > seeking to the end the span

Is this method actually seeking to the end of the span?

Sources/CacheAdvance/DataExtensions.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    /// The width of an encoded byte count.
+    static let bytesStorageLength = MemoryLayout<Bytes>.size
+
+    /// The width of an encoded message span.
+    static let messageSpanStorageLength = MemoryLayout<MessageSpan>.size
+
+    /// A marker written at the end of the newest message written to disk.
```
bachand: I wonder if we should explicitly clarify that this works because we will never write message of 0 length? (assuming I am understanding correctly)

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        try? FileManager.default.removeItem(at: testFileLocation)
+    }
+
+    // MARK: Behavior Tests
+
+    func test_append_singleMessageThatFitsCanBeRetrieved() throws {
```
bachand: I'd try to break this method name up a bit more.
Maybe something like: `test_append_singleMessageThatFits_canBeRetrieved()`
Then the next one can be: `test_append_singleMessageThatDoesNotFit_fails()`

I generally think of naming my tests methods in this format:

`test_<methodName>_<context>_<expectedBehavior>()`

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -0,0 +1,391 @@
...
+
+    func test_append_singleMessageFailsIfDoesNotFit() throws {
+        let message = "This is a test"
+        let cache = try CacheAdvance<String>(
+            file: testFileLocation,
+            maximumBytes: try requiredByteCount(for: [message], cacheWillRoll: false) - 1,
+            shouldOverwriteOldMessages: false)
+
+        XCTAssertThrowsError(try cache.append(message: message)) {
+            XCTAssertEqual($0 as? CacheAdvanceWriteError, CacheAdvanceWriteError.messageDataTooLarge)
```
bachand: Nice, I like that we're checking the error type.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -0,0 +1,391 @@
...
+    func test_append_dropsFirstMessageIfCacheRollsAndLastMessageDoesNotFitAndIsShorterThanFirstMessage() throws {
+        let cache = try CacheAdvance<String>(
+            file: testFileLocation,
+            maximumBytes: try requiredByteCount(for: lorumIpsumMessages, cacheWillRoll: true),
+            shouldOverwriteOldMessages: true)
+        for message in lorumIpsumMessages {
+            try cache.append(message: message)
+        }
+
+        // Append a message that is shorter than the first message in lorumIpsumMessages.
```
bachand: Can we turn this comment into an `XCTAssert` or some sort?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -0,0 +1,391 @@
...
+    func test_append_dropsFirstTwoMessagesIfCacheRollsAndLastMessageDoesNotFitAndIsLargerThanFirstMessage() throws {
+        let cache = try CacheAdvance<String>(
+            file: testFileLocation,
+            maximumBytes: try requiredByteCount(for: lorumIpsumMessages, cacheWillRoll: true),
+            shouldOverwriteOldMessages: true)
+        for message in lorumIpsumMessages {
+            try cache.append(message: message)
+        }
+
+        // Append a message that is slightly longer than the first message in lorumIpsumMessages.
```
bachand: Turn into `XCTAssert` of some sort?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -0,0 +1,391 @@
...
+        "In ultricies risus vel nisi convallis fermentum.",
+        "Etiam tempor nisi nulla,",
+        "eu pulvinar nisl pretium ut.",
+        "Cras ullamcorper enim nisl,",
+        "at tempus arcu sagittis quis.",
+    ]
+
+    private func requiredByteCount<T: Codable>(for messages: [T], cacheWillRoll: Bool) throws -> UInt64 {
+        let encoder = JSONEncoder()
+        let messageSpanSuffixLength = cacheWillRoll ? Bytes(Data.messageSpanStorageLength + Data.bytesStorageLength) : Bytes(Data.messageSpanStorageLength)
```
bachand: It could be cool if there was one or more `internal` methods we could call on the production code to compute this stuff, rather than doing it ourself here. But not a big deal. Just a thought.

Sources/CacheAdvance/DataExtensions.swift
bachand: I like the delimiter being 0 👍

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: objective-c
```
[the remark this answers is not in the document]
dfed: `swift` is relatively new to travis (to the point that I didn't know it existed!). It's currently just an alias: https://docs.travis-ci.com/user/languages/objective-c/#objective-c-vs-swift. Seems like a worthy change though

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+install:
+  - bundle install
+  - brew outdated carthage || brew upgrade carthage
```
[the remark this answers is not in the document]
dfed: Not yet, but I will. Downside of copy/pasting over configs from other repos. This will help when I do eventually support Carthage (coming soon!)

Scripts/ci.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -l
+set -ex
```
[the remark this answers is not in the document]
dfed: exits the script on any non-zero exit code

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+script:
+  - ./Scripts/ci.sh
+matrix:
+  include:
+  - osx_image: xcode11.2
+    env: ACTION="swift-package";SDK="iphonesimulator";DESTINATION="platform=iOS Simulator,OS=13.2.2,name=iPad Pro (12.9-inch) (3rd generation)";SHOULD_TEST="true"
```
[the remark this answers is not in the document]
dfed: These environment variables are utilized in my [ci.sh](https://github.com/dfed/CacheAdvance/pull/1/files#diff-f7ffaea5d0508d52299b778bce4ed600R4) script. The script makes it easy to run CI with different matrix input.

In terms of how I figured out what the DESTINATION, SDK, etc to use... honestly a lot of getting it wrong and reading errors in CI (and `xcodebuild`'s `help` page) until I figured out a combo that worked. I now use the same configuration on all of my repos that are built for multiple deployment targets.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - osx_image: xcode11.2
+    env: ACTION="swift-package";SDK="appletvsimulator13.2";DESTINATION="platform=tvOS Simulator,name=Apple TV";SHOULD_TEST="true"
+  - osx_image: xcode11.2
+    # We can't run tests on macOS because CacheAdvance supports a minimum of macOS 10.15, and Travis CI's machines run on macOS 10.14.
```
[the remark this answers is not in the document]
dfed: Yes (I plan on filing a bunch of issues when I merge 😅). I'll add this to the issue, but the two options I see are:
1. Move to Circle CI (which supports Catalina)
2. Wait until travis supports Catalina (it can't be stuck on 10.14 forever).

Contributing.md
```diff
@@ -0,0 +1,15 @@
...
+
+Keep your Pull Requests small. Small PRs are easier to reason about which makes them significantly more likely to get merged.
+
+### Issues before features
+
+If you want to add a feature, please file an [Issue](../../issues) first. An Issue gives us the opportunity to discuss the requirements and implications of a feature with you before you start writing code.
```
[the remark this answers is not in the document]
dfed: It is possible!

https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository

Good idea. I'll add this to my to-do list.

Scripts/build.swift
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env swift
+
+import Foundation
+
+// Usage: ci.swift sdk destination should_test
```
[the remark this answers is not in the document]
dfed: I'm backing these changes out. The way I'm building the argument list in `Process` isn't working the way I want (and it's causing a CI failure which I can't reproduce locally, which makes me think the failure is related to macOS 10.14). I'll pull this change into a separate PR. The Swift -> shell interface is pretty gnarly (but hopefully more maintainable once I get it working).

Scripts/build.swift
```diff
@@ -0,0 +1,52 @@
...
+]
+if !destination.isEmpty {
+    xcodeBuildArguments.append("-destination=\(destination)")
+}
+xcodeBuildArguments.append("build")
+if shouldTest {
+    xcodeBuildArguments.append("test")
+}
+
+try execute(commandPath: "/usr/bin/xcodebuild", arguments: xcodeBuildArguments)
```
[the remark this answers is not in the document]
dfed: That's the hope!

Scripts/ci.sh
```diff
@@ -0,0 +1,22 @@
...
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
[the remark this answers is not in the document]
dfed: Yeah I'll add a comment. Good call.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+    // MARK: Initialization
+
+    /// Creates a new instance of the receiver.
+    ///
+    /// - Parameters:
+    ///   - file: The file URL indicating the desired location of the on-disk store. This file should already exist.
+    ///   - maximumBytes: The maximum size of the cache, in bytes. Logs larger than this size will fail to append to the store.
+    ///   - shouldOverwriteOldMessages: When `true`, once the on-disk store exceeds maximumBytes, new entries will replace the oldest entry.
+    ///
+    /// - Warning: `maximumBytes` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache leads to undefined behavior.
```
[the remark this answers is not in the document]
dfed: I'll file this as an issue after merge. Thanks for writing this out 🙂

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+        self.maximumBytes = maximumBytes
+        self.shouldOverwriteOldMessages = shouldOverwriteOldMessages
+
+        writer = try FileHandle(forWritingTo: file)
+        reader = try FileHandle(forReadingFrom: file)
+
+        lengthOfMessageSuffix = {
+            if shouldOverwriteOldMessages {
+                /// The message suffix requires space for both `endOfNewestMessageMarker` and `offsetOfFirstMessage` after each message when rolling is enabled.
+                return Bytes(Data.endOfNewestMessageMarker.count) + Bytes(Data.bytesStorageLength)
```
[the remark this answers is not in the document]
dfed: I like the idea of an enum. I'll break that out as a separate PR.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+
+        guard messageLength > 0 else {
+            // The message length has the same value as as our `endOfNewestMessageMarker`.
+            // Storing this message could cause data corruption by fooling the cache into thinking the prior message is the last in the cache.
+            // JSON values (even empty ones) are always encoded with length > 0. If this condition is hit, the message is clearly corrupt.
+            throw CacheAdvanceWriteError.messageCorrupted
+        }
+
+        let bytesNeededToStoreMessage = messageLength + lengthOfMessageSuffix
+        guard bytesNeededToStoreMessage <= maximumBytes && bytesNeededToStoreMessage < Bytes(MessageSpan.max) else {
```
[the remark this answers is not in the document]
dfed: > It seems like we'd never want to be in a situation where `MessageSpan.max` is greater than `maximumBytes`.

I think you mean the inverse? `MessageSpan.max` is `UInt32.max` (i.e. four gigabytes), and `maximumBytes` is the maximum byte size of this cache, which is likely to be pretty small by comparison.

`MessageSpan` is an `UInt32` and `Bytes` is `UInt64`. I wanted to enable _massive_ maximum caches for those that wanted it, which means that it is entirely possible for `maximumBytes` to greatly exceed a `MessageSpan`'s capability. It is currently user error to log a single message larger than four gigabytes. Using a `UInt32` to define the `MessageSpan` allows me to store the length of each message in half the space vs the space if I'd used an `UInt64`. Martin suggested earlier in this PR that I make `MessageSpan` a generic so that folk can write messages longer than `UInt32.max` in length if they want to opt-into doing so. I think we can tackle that kind of customizability once we have a header or file metadata stored somewhere.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+            // The message is too long to be written to a cache of this size.
+            throw CacheAdvanceWriteError.messageDataTooLarge
+        }
+
+        let cacheHasSpaceForNewMessageBeforeEndOfFile = writer.offsetInFile + bytesNeededToStoreMessage <= maximumBytes
```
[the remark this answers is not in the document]
dfed: I can write a file that **is** `maximumBytes` long.

If I make this a `<` rather than a `<=`, my unit tests that try to write a message to a file that is the same length as `maximumBytes` fails.

I'm reasonably confident I don't have an off-by-one error here... unless I've managed to make the same off-by-one error in both my tests and the production code. 0-indexing is hard 😅.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+                try writer.truncate(atOffset: writer.offsetInFile)
+
+                // Set the offset back to the beginning of the file.
+                try writer.seek(toOffset: 0)
+
+                // We know the oldest message is at the beginning of the file, since we just tossed out the rest of the file.
+                try reader.seek(toOffset: 0)
+
+                // We know we're about to overwrite the oldest message, so advance the reader to the second oldest message.
+                try reader.seekToNextMessage(shouldSeekToOldestMessageIfFound: false, cacheOverwritesOldMessages: true)
```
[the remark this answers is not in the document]
dfed: It is doing a lot, for sure. None of this logic is reused though, so keeping the logic co-located felt helpful?

I could probably factor out the `shouldOverwriteOldMessages` if-block to its own method, but I'm not sure how much benefit I'd get from that.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+            throw CacheAdvanceWriteError.messageDataTooLarge
+        }
+    }
+
+    /// Fetches all messages from the cache.
+    public func messages() throws -> [T] {
+        try setUpFileHandlesIfNecessary()
+
+        var messages = [T]()
+        while let encodedMessage = try reader.nextEncodedMessage(cacheOverwritesOldMessages: shouldOverwriteOldMessages) {
```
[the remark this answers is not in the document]
dfed: > It seems like the reader is stateful.

It is! At the end of each `public` method call, the `reader` is at the beginning of the oldest known message.

> how is the reader being reset?

Because we read until the reader is back on the oldest known message 🙂

> invoking this twice in a row could be a good test

can't hurt! I'll add it.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+
+        // Now that we know where to write, we need to figure out where the oldest message is.
+        try reader.seekToBeginningOfOldestMessage(cacheOverwritesOldMessages: shouldOverwriteOldMessages)
+
+        hasSetUpFileHandles = true
+    }
+
+    /// Advances the reader until there is room to store a new message without writing past the reader.
+    /// - Parameter messageLength: the length of the next message that will be written.
+    private func prepareReaderForWriting(dataOfLength messageLength: Bytes) throws {
```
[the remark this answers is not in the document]
dfed: I'll add documentation. The reader is basically useless (outside of reading all messages) in a cache that doesn't overwrite old messages.

I'm musing here, but I'm curious how much benefit we're getting from having a single cache type that can either overwrite old messages and one that throws when it is full in a single class.

Maybe it's worth splitting out the functionality into a `CacheAdvance` and an `OverwritingCacheAdvance`? (This would be a separate PR)

Sources/CacheAdvance/CacheAdvance.swift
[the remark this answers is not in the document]
dfed: I think this comment should move to the public method where we do the check. This method is very much a utility – it doesn't do any sanity checking.

Sources/CacheAdvance/CacheAdvance.swift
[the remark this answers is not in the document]
dfed: We are 🙂

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -0,0 +1,215 @@
...
+    /// By the time this method returns, the `writer`'s `offsetInFile` is always set to the beginning of the written `endOfNewestMessageMarker`, such that the next message written will overwrite the marker.
+    ///
+    /// - Parameters:
+    ///   - messageData: an `EncodableMessage`'s encoded data. Must be smaller than both `maximumBytes` and `MessageSpan.max`.
+    ///   - offsetInFileOfOldestMessage: An optional big-endian representation of the offset in the file where the oldest message begins.
+    ///                                  This data is should only be included in caches that overwrite their oldest messages.
+    private func write(messageData: Data, offsetInFileOfOldestMessage: Data?) throws {
+        // Create data to write from combined message and suffix.
+        var dataToWrite = messageData + Data.endOfNewestMessageMarker
+        if let offsetInFileOfOldestMessage = offsetInFileOfOldestMessage, offsetInFileOfOldestMessage.count > 0 {
```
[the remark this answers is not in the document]
dfed: Good call-out. I think the `offsetInFileOfOldestMessage.count > 0` check is superfluous. If it's empty... then appending it doesn't matter.

Sources/CacheAdvance/CacheAdvanceErrors.swift
dfed: We are writing atomically. But it's still possible for an application to crash while we're writing (trivial example: while we're writing, the app divides by zero on another thread)

Sources/CacheAdvance/DataExtensions.swift
dfed: I like the thinking. I'll consider breaking this out in a separate PR.

Sources/CacheAdvance/CacheAdvance.swift
dfed: The above scenario will lead to data loss. Byte 0 is always the beginning of `MessageSpan` followed by an encoded Message, so we'll never read garbage data. What will happen is that we'll fail to read the messages from the oldest message in the file until the end of the file, if the oldest message isn't at Byte 0.

I'll add a warning for this.

Sources/CacheAdvance/EncodableMessage.swift
dfed: yup! Super handy

Sources/CacheAdvance/EncodableMessage.swift
dfed: almost certainly. Though Apple's APIs for doing so are more prone to crashing (rather than `throw`ing). Worth investigating though. I'll add this as an issue after merge.

Sources/CacheAdvance/FileHandleExtensions.swift
dfed: The only way we can get an `emptyRead` is if:
```
// We haven't written anything to this file yet, or we've reached the end of the file.
```

I think given that comment I can simplify this code.

Sources/CacheAdvance/FileHandleExtensions.swift
dfed: since `seek` takes a `UInt64`, it seemed more semantic to use that type here.

Imagine a world where we decide to only support a 4GB file and change `Bytes` to be a typealias for `UInt32. In that world, we wouldn't want this line to fail.

Sources/CacheAdvance/FileHandleExtensions.swift
dfed: yup! `__readDataUp` reads and seeks.

Sources/CacheAdvance/FileHandleExtensions.swift
dfed: great call. updated!

Sources/CacheAdvance/DataExtensions.swift
dfed: You're understanding correctly. I have a comment that captures this truth elsewhere, but it can't hurt to mention that here as well.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: Would certainly allow for faster iteration on new file formats

I’ll likely improve this the moment I want to change the file format 😂

In the meantime I’ll save this for a later PR

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: I think the name of the method should help enough. There’s only one assertion per test (with the exception of the methods expecting errors to be thrown), so I’m not sure how much an assertion comment helps.

Commenting explicitly what this line is doing I think does help with the long-term maintenance of this test.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: I don’t normally split out context... but I like it.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -0,0 +1,435 @@
...
+        let cache = try CacheAdvance<String>(
+            file: testFileLocation,
+            maximumBytes: 50,
+            shouldOverwriteOldMessages: true)
+
+        let messages = try cache.messages()
+        XCTAssertEqual(messages, [])
+    }
+
+    func test_append_singleMessageThatFitsCanBeRetrieved() throws {
```
dfed:
```suggestion
    func test_append_singleMessageThatFits_canBeRetrieved() throws {
```

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -0,0 +1,435 @@
...
+            file: testFileLocation,
+            maximumBytes: try requiredByteCount(for: [message], cacheWillOverwriteOldestMessages: false),
+            shouldOverwriteOldMessages: false)
+        try cache.append(message: message)
+
+        let messages = try cache.messages()
+        XCTAssertEqual(messages, [message])
+    }
+
+    func test_append_singleMessageFailsIfDoesNotFit() throws {
```
dfed:
```suggestion
    func test_append_singleMessageThatDoesNotFit_throwsError() throws {
```

Sources/CacheAdvance/Bytes.swift
```diff
@@ -0,0 +1,55 @@
...
+        }
+        self = NSSwapBigBytesToHost(decodedSize.load(as: Bytes.self))
+    }
+
+    /// Converts megabytes into the receiver. Will never overflow.
+    ///
+    /// - Parameter megabytes: The number of megabytes to convert.
+    ///
+    /// - Note: `megabytes` are converted to `bytes` by multiplying by 1,000,000.
+    ///         This conversion matches how Apple's systems calcuate `bytes`.
```
dfed:
```suggestion
    ///         This conversion matches how Apple's systems measure megabytes.
```

Sources/CacheAdvance/FileHandleExtensions.swift
dfed: Looking at this again, I think you're right. In a cache that overrides old messages, we'll hit the oldest message offset marker before we have an empty read – getting an `emptyRead` in that case would likely mean there was some kind of programmer error (we'd have to seek to the next message span after reading the oldest message offset marker). I'll clarify with a comment.

dfed:
> I'd feel more comfortable merging if we wrote tests for the file handle extensions.

These are indirectly tested (with high code coverage) through the existing unit tests. I agree it'd be better to have explicit tests here, but We’ll Get There™ in a future PR.

Sources/CacheAdvance/EncodableMessage.swift
dfed: I just did an investigation. It looks like the binary property list is much less space efficient than a straight JSON encoding.

The reason (I think, I'd need to investigate further) is that I store messages one at a time, and the overhead of storing a list per message outweighs whatever benefits the binary property list gives us.

The following statements resolve to `true`:
```
try PropertyListSerialization.data(fromPropertyList: [], format: .binary, options: 0).count == 42
try JSONEncoder().encode(Data()).count == 2
```

bachand: @dfed this makes some sense to me, since property lists have a header section and more overhead than JSON. What happens if we write a meaningful amount of data, eg 50 dictionaries with a few keys each?

dfed: Surprisingly, encoding the entire `lorumIpsumMessages` array had the same result. Property list serialization is larger than json encoding.

```
try PropertyListSerialization.data(fromPropertyList: lorumIpsumMessages, format: .binary, options: 0).count == 7463
try JSONEncoder().encode(lorumIpsumMessages).count == 6822
```

I wonder if the JSONEncoder starts losing ground if each message has multiple fields. Right now we're only testing strings.

bachand: Wow, good learnings here. Not what I expected. Thanks for checking!

dfed: Just tried encoding an absurdly large dictionary, with each entry having hundreds of key:value pairs, and JSONEncoder still won out on size.

dfed: Not what I expected either! Thanks for prompting me to check 🙂

bachand: This likely won’t matter but I do want to point out there is a cleaner way to do this using `Codable`
https://developer.apple.com/documentation/foundation/propertylistencoder

dfed: Good point! Just tried it, and:
```
try PropertyListEncoder().encode(lorumIpsumMessages).count == PropertyListSerialization.data(fromPropertyList: lorumIpsumMessages, format: .binary, options: 0).count
```

Issue #4, opened 2019-12-02
Add Cocoapods support


dfed:
We should have a podspec, and lint it in CI

PR #12, opened 2019-12-02
Support Cocoapods

dfed:
This PR resolves https://github.com/dfed/CacheAdvance/issues/4


Scripts/ci.sh
```diff
@@ -24,10 +24,7 @@ if [ $ACTION == "swift-package" ]; then
 fi
 
 if [ $ACTION == "pod-lint" ]; then
-  swift package generate-xcodeproj --output generated/
```
dfed: the generated project isn't needed to validate a pod

PR #15, opened 2019-12-04
Create a better namespace for extensions on Data

dfed:
This PR is a follow-up to https://github.com/dfed/CacheAdvance/pull/1#discussion_r352308048. `Data` was an odd namespace for the values I was placing on it.

I was also repeating some logic between `Bytes` and `MessageSpan`. In this PR, I consolidate that logic on `protocol EncodableInteger`. Using a shared protocol also moves us one step closer to a world where we could support enabling the consumer of a `CacheAdvance` to specify the underlying type of `MessageSpan` and `Bytes` via generics. It also simplifies the code a bit.

cc @bachand @martinmroz


Sources/CacheAdvance/EncodableInteger.swift
```diff
@@ -0,0 +1,61 @@
...
+import Foundation
+
+/// A protocol representing an integer that can be encoded as data.
+protocol EncodableInteger {
+
+    /// Converts the big-endian value in x to the current endian format and returns the resulting value.
+    static func swapBigIntToHost(_ x: Self) -> Self
+
+    /// The big-endian representation of self.
+    @inlinable var bigEndian: Self { get }
```
bachand: Why add `@inlinable`?

Tests/CacheAdvanceTests/EncodableIntegerTests.swift
```diff
@@ -0,0 +1,36 @@
...
+    // MARK: Behavior Tests
+
+    func test_init_canBeInitializedFromEncodedDta() {
+        let expectedValue: UInt64 = 10
+        XCTAssertEqual(UInt64(Data(expectedValue)), expectedValue)
+    }
+
+    func test_init_returnsNilWhenInitializedFromDataOfIncorrectLength() {
+        var expectedValue: UInt64 = 10
+        XCTAssertEqual(UInt64(Data(bytes: &expectedValue, count: UInt64.storageLength + 1)), nil)
```
bachand: `XCTAssertNil`

Sources/CacheAdvance/EncodableInteger.swift
dfed: The property is `@inlinable` already on `UInt32` and `UInt64` (and likely other integer types). Figured staying consistent was valuable.

Tests/CacheAdvanceTests/EncodableIntegerTests.swift
dfed:
```suggestion
        XCTAssertNil(UInt64(Data(bytes: &expectedValue, count: UInt64.storageLength + 1)))
```

Sources/CacheAdvance/EncodableInteger.swift
```diff
@@ -0,0 +1,61 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// A protocol representing an integer that can be encoded as data.
+protocol EncodableInteger {
```
fdiaz: I'm not sure I follow this. The name [encodable](https://developer.apple.com/documentation/swift/encodable) [integer](https://developer.apple.com/documentation/swift/int) makes it feel like you're "making an integer encodable" (which we know this is always true).

Then in the description you talk about being able to represent an Int as Data, but there's no reference to any Data in the protocol signature.

From what I can tell this protocol is mostly to convert to manipulate the big-endian of an instance?

fdiaz: Should this protocol be bound to [ExpressibleByIntegerLiteral](https://developer.apple.com/documentation/swift/expressiblebyintegerliteral)?

dfed: This protocol defines a type that:
- Can be encoded as `Data` of width `storageLength`
- Can be initialized back from `Data` of width `storageLength`
- Can be encoded in `bigEndian` format and read back out in the host's endian preference.

To be fair, any `integer` can do this, but this protocol allows me to add conveniences. The good news is that it is an `internal` protocol (for now). I'm open to different naming given the above.

Also, I 100% agree this should be `ExpressibleByIntegerLiteral`. I'd forgotten that existed.

bachand: Huh, how does this compile when we bind this protocol to `ExpressibleByIntegerLiteral` if neither `UInt32` or `UInt64` are listed as conforming types? What am I missing?
AC5F3721-B8C5-4671-A85E-E796D7BC544B

dfed: The key is `Numeric`, which is a protocol that inherits from this protocol and also is conformed to (indirectly) by `UInt32` and `UInt64`

dfed: Okay I think I can simplify the naming by using `FixedWidthInteger`. Commit incoming.

dfed: https://github.com/dfed/CacheAdvance/pull/15/commits/f6475da880df78eb74fdecd09a52b990d7519bc6 should make the naming a bit more clear

bachand: Yep! Wow that inheritance chain goes deep!

bachand:
Nice cleanup!

Sources/CacheAdvance/EncodableInteger.swift
dfed: Indeed! Thanks for asking. Wouldn’t have found `FixedWidthInteger` if you hadn’t

bachand:
> FWIW I was suggesting doing
> 
> `protocol BigEndianHostSwappable where Self: FixedWidthInteger`
> 
> which controls who can conform to this (which is what I think we want here?) instead of requiring the protocol to also conform to `FixedWidthInteger`.

@fdiaz what is the practical difference between `BigEndianHostSwappable: FixedWidthInteger` and `BigEndianHostSwappable where Self: FixedWidthInteger`? Asking to learn.

dfed:
@fdiaz that's a good call. I'll put out a PR.

@bachand when we have `protocol BigEndianHostSwappable: FixedWidthInteger`, someone can create an empty object that conforms to `BigEndianHostSwappable`, and the compiler will guide them to create methods and properties such that they conform to both `BigEndianHostSwappable` and `FixedWidthInteger`.

When we have `protocol BigEndianHostSwappable where Self: FixedWidthInteger`, someone can create an empty object that conforms to `BigEndianHostSwappable`, and the compiler will tell them that the object _does not conform_ to `FixedWidthInteger`.

In the latter case, we're saying that `BigEndianHostSwappable` applies to objects that independently are `FixedWidthInteger`. In the former case, we're saying that to conform to `BigEndianHostSwappable` you also have to implement all of the fields and properties on `FixedWidthInteger`.

The practical difference, as far as I'm aware, is the error the compiler gives you. Semantically, the former is stating the intent, which is that `BigEndianHostSwappable` is a decoration on objects that conform to `FixedWidthInteger`.

bachand:
Thanks for the explanation. That's helpful.

Issue #9, opened 2019-12-02
Support macOS 10.14, iOS 12, tvOS 12, and watchOS 5


dfed:
To do this, we'll need to write wrappers for the FileHandle APIs we are using that were introduced in macOS 10.15, iOS 13, tvOS 13, and watchOS 6.

The older methods raise Objective-C exceptions rather than `throw`ing catchable Swift `Error`s, so we'll need to write the wrappers in Objective-C.

Swift package manager does not support mixed-language modules, so we'll need to create another repo for the Objective-C wrappers.

Once we have the Objective-C wrappers, lowering our supported version should be as simple as importing the new module, lowering our supported version, and adding the newly supported versions to our CI matrix.

dfed:
We may be able to utilize https://github.com/williamFalcon/SwiftTryCatch for the Objective-C wrapper.

The tricky bit will be getting the dependency working in Swift Package Manager (SPM), Cocoapods, and Carthage. I'll need to fork the repo to get SPM working. Cocoapods has transitive dependencies. I'm not sure about Carthage (need to do more research).

dfed:
This was resolved in #19

PR #19, opened 2019-12-20
Support macOS 10.14, iOS 12, tvOS 12, and watchOS 5

dfed:
This PR addresses https://github.com/dfed/CacheAdvance/issues/9

I used code from https://github.com/williamFalcon/SwiftTryCatch to implement the Objective-C code that would enable Swift to catch exceptions raised by Objective-C. That repo's podspec doesn't list tvOS, macOS, or watchOS as supported platforms, so I wasn't able to use Cocoapods to bring in this dependency. To simplify the dependency story, I copied over the files I needed from `williamFalcon`'s repo and modified them as needed. I also wrote unit tests for the code I brought over.

Please note that this PR is not as big as it looks! There's a couple mass renames that inflates the line count.


CacheAdvance.podspec
```diff
@@ -1,15 +1,15 @@
...
   s.homepage = 'https://github.com/dfed/CacheAdvance'
   s.authors  = 'Dan Federman'
   s.source   = { :git => 'https://github.com/dfed/CacheAdvance.git', :tag => s.version }
   s.swift_version = '5.1'
-  s.source_files = 'Sources/CacheAdvance/**/*.{swift}'
-  s.ios.deployment_target = '13.0'
-  s.tvos.deployment_target = '13.0'
-  s.watchos.deployment_target = '6.0'
-  s.macos.deployment_target = '10.15'
+  s.source_files = 'Sources/CacheAdvance/**/*.{swift}', 'Sources/SwiftTryCatch/**/*.{h,m}'
```
dfed: Because there is no viable `SwiftTryCatch` pod, I needed to include the source files for this code directly in the `CacheAdvance` pod. Note that I'm not exporting any headers in the podspec, so this should be an implementation detail of the pod.

Package.swift
```diff
@@ -19,11 +19,22 @@ let package = Package(
     targets: [
         .target(
             name: "CacheAdvance",
-            dependencies: []),
+            dependencies: ["SwiftTryCatch"],
+            swiftSettings: [.define("SWIFT_PACKAGE_MANAGER")]
```
dfed: This enables the code to tell the difference between a build from SPM and a build from Cocoapods. In Cocoapods, `SwiftTryCatch` is not a separate module. In SPM, `SwiftTryCatch` _has_ to be a separate target (since SPM doesn't allow mixed-language targets). So I only `import SwiftTryCatch` when `SWIFT_PACKAGE_MANAGER` is set.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -57,8 +57,8 @@ public final class CacheAdvance<T: Codable> {
     }
 
     deinit {
-        try? writer.close()
-        try? reader.close()
+        try? writer.closeHandle()
+        try? reader.closeHandle()
```
dfed: these method renames are due to code that is below. No functional changes. Keep scrolling!

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -21,12 +21,57 @@ extension FileHandle {
...
+    func write(data: Data) throws {
+        if #available(iOS 13.0, tvOS 13.0, watchOS 6.0, OSX 15.0, *) {
+            return try __write(data, error: ())
+        } else {
+            return try ObjectiveC.unsafe { write(data) }
+        }
+    }
+
+    /// A method to seek on a file handle that is safe to call in Swift from any operation system version.
+    func seek(to offset: UInt64) throws {
```
dfed: I don't love this method name. I'm open to suggestions.

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -21,12 +21,57 @@ extension FileHandle {
...
+    func closeHandle() throws {
+        if #available(iOS 13.0, tvOS 13.0, watchOS 6.0, OSX 15.0, *) {
+            return try close()
+        } else {
+            return try ObjectiveC.unsafe { closeFile() }
+        }
+    }
+
+    /// A method to truncate a file handle that is safe to call in Swift from any operation system version.
+    func truncate(at offset: UInt64) throws {
```
dfed: I don't love this method name. I'm open to suggestions.

Sources/SwiftTryCatch/SwiftTryCatch.h
```diff
@@ -0,0 +1,33 @@
+//
+//  SwiftTryCatch.h
+//
+//  Created by William Falcon on 10/10/14.
```
dfed: This code was adapted from https://github.com/williamFalcon/SwiftTryCatch/

Sources/SwiftTryCatch/SwiftTryCatch.m
```diff
@@ -0,0 +1,42 @@
+//
+//  SwiftTryCatch.h
+//
+//  Created by William Falcon on 10/10/14.
```
dfed: This code was adapted from This code was adapted from https://github.com/williamFalcon/SwiftTryCatch/

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -37,7 +37,7 @@ final class CacheAdvanceTests: XCTestCase {
     // MARK: Behavior Tests
 
     func test_messages_canReadEmptyCacheThatDoesNotOverwriteOldestMessages() throws {
-        let cache = try CacheAdvance<String>(
+        let cache = try CacheAdvance<TestableMessage>(
```
dfed: In order to get tests passing on the older platforms, I needed to encode an object other than `String`. Trying to encode the `String` would lead to the error:
> The data couldn’t be written because it isn’t in the correct format.

You can see a sample of this error (before I figured out how to fix it 😅) in [this build](https://travis-ci.com/dfed/CacheAdvance/jobs/269437929#L1531).

.travis.yml
```diff
@@ -7,17 +7,23 @@ script:
 matrix:
   include:
   - osx_image: xcode11.2
-    env: ACTION="swift-package";SDK="iphonesimulator";DESTINATION="platform=iOS Simulator,OS=13.2.2,name=iPad Pro (12.9-inch) (3rd generation)";SHOULD_TEST="true"
+    env: ACTION="swift-package";PLATFORMS="iOS12,iOS13";
+
   - osx_image: xcode11.2
-    env: ACTION="swift-package";SDK="appletvsimulator13.2";DESTINATION="platform=tvOS Simulator,name=Apple TV";SHOULD_TEST="true";ENABLE_CODE_COVERAGE="true"
+    env: ACTION="swift-package";PLATFORMS="tvOS12,tvOS13";
```
dfed: In order to not drop code coverage, I need to run tests on multiple platforms before uploading code coverage numbers

.travis.yml
```diff
@@ -7,17 +7,23 @@ script:
 matrix:
   include:
   - osx_image: xcode11.2
-    env: ACTION="swift-package";SDK="iphonesimulator";DESTINATION="platform=iOS Simulator,OS=13.2.2,name=iPad Pro (12.9-inch) (3rd generation)";SHOULD_TEST="true"
+    env: ACTION="swift-package";PLATFORMS="iOS_12,iOS_13";
+
   - osx_image: xcode11.2
-    env: ACTION="swift-package";SDK="appletvsimulator13.2";DESTINATION="platform=tvOS Simulator,name=Apple TV";SHOULD_TEST="true";ENABLE_CODE_COVERAGE="true"
+    env: ACTION="swift-package";PLATFORMS="tvOS_12,tvOS_13";
```
dfed: In order to not drop code coverage, I need to run tests on multiple platforms before uploading code coverage numbers

Scripts/build.swift
```diff
@@ -20,38 +20,138 @@ enum TaskError: Error {
...
+    case iOS_12
+    case iOS_13
+    case tvOS_12
+    case tvOS_13
+    case macOS_10_14
+    case macOS_10_15
+    case watchOS_5
+    case watchOS_6
+
+    var destination: String {
```
dfed: these values were previously in `ci.sh`

.travis.yml
```diff
@@ -7,17 +7,25 @@ script:
...
-    env: ACTION="swift-package";SDK="appletvsimulator13.2";DESTINATION="platform=tvOS Simulator,name=Apple TV";SHOULD_TEST="true";ENABLE_CODE_COVERAGE="true"
+    env: ACTION="swift-package";PLATFORMS="tvOS_12";
     after_success:
       bash <(curl -s https://codecov.io/bash) -J '^CacheAdvance$' -D .build/derivedData -t 8344b011-6b2a-4b3d-a573-eaf49684318e
   - osx_image: xcode11.2
-    # We can't run tests on macOS because CacheAdvance supports a minimum of macOS 10.15, and Travis CI's machines run on macOS 10.14.
-    env: ACTION="swift-package";SDK="macosx10.15";DESTINATION="platform=OS X";SHOULD_TEST="false"
+    env: ACTION="swift-package";PLATFORMS="tvOS_13";
+    after_success:
+      bash <(curl -s https://codecov.io/bash) -J '^CacheAdvance$' -D .build/derivedData -t 8344b011-6b2a-4b3d-a573-eaf49684318e
```
dfed: Codecov is smart enough to merge the reports

.travis.yml
```diff
@@ -7,17 +7,25 @@ script:
 matrix:
   include:
   - osx_image: xcode11.2
-    env: ACTION="swift-package";SDK="iphonesimulator";DESTINATION="platform=iOS Simulator,OS=13.2.2,name=iPad Pro (12.9-inch) (3rd generation)";SHOULD_TEST="true"
+    env: ACTION="swift-package";PLATFORMS="iOS_12,iOS_13";
+
   - osx_image: xcode11.2
-    env: ACTION="swift-package";SDK="appletvsimulator13.2";DESTINATION="platform=tvOS Simulator,name=Apple TV";SHOULD_TEST="true";ENABLE_CODE_COVERAGE="true"
```
dfed: While I was here I simplified the configuration in this file, and pushed the complexity to the easier to read `build.swift` file. (I originally thought I had to change this system to get code coverage reports merged, but it turns out I didn't 🤦‍♂️)

dfed:
cc @bachand this is now ready for review (I'd add you as a reviewer, but you've got to accept my "collaborator" invite first 😉)

Scripts/build.swift
```diff
@@ -20,38 +20,132 @@ enum TaskError: Error {
...
 }
 
-guard CommandLine.arguments.count > 3 else {
-    print("Usage: build.swift sdk destination should_test? enable_code_coverage?")
+enum Platform: String, CaseIterable, CustomStringConvertible {
+    case iOS_12
+    case iOS_13
+    case tvOS_12
+    case tvOS_13
+    case macOS_10_15
```
dfed: Note that we don't have a `macOS_10_14`, since this project requires Xcode 11 to build, and there is no CI configuration on Travis CI that uses macOS 10.14's SDK but enables building with Xcode 11.

Tests/SwiftTryCatchTests/SwiftTryCatchTests.swift
```diff
@@ -0,0 +1,74 @@
...
+
+    func test_try_stopsExecutingOnRaise() {
+        let doomedFileHandle = FileHandle(fileDescriptor: 101)
+        var didExecuteAfterRaisedException = false
+        SwiftTryCatch.try({
+            // This will raise an exception.
+            doomedFileHandle.readData(ofLength: 5)
+            didExecuteAfterRaisedException = false
+        }, catch: { _ in })
+        XCTAssertFalse(didExecuteAfterRaisedException)
```
fdiaz: This tests is not ideal since we set `didExecuteAfterRaisedException` as false initally, and then we go to a closure, so in theory if this closure were to become async we would not get in the closure and this test will still succeed.

Also, we're setting `didExecuteAfterRaisedException = false` (even though it starts as false) in line 44, so we're actually not modifying anything and then checking if the original value is still set.

I'd suggest something like this instead:

```swift
let doomedFileHandle = FileHandle(fileDescriptor: 101)
var didStopAfterRaisedException = false
SwiftTryCatch.try({
    didStopAfterRaisedException = true
    // This will raise an exception.
    doomedFileHandle.readData(ofLength: 5)
    didStopAfterRaisedException = false
}, catch: { _ in })

XCTAssertTrue(didStopAfterRaisedException)
```

Package.swift
```diff
@@ -19,11 +19,21 @@ let package = Package(
...
+        ),
         .testTarget(
             name: "CacheAdvanceTests",
-            dependencies: ["CacheAdvance"])
+            dependencies: ["CacheAdvance"]),
+        .target(
+            name: "SwiftTryCatch",
+            dependencies: [],
+            publicHeadersPath: "./",
+            swiftSettings: [SwiftSetting.define("-fobjc-arc-exceptions")]
```
fdiaz: Maybe we should add a comment here explaining why we need this?

Sources/CacheAdvance/FileHandleExtensions.swift
fdiaz: this should probably be `seek(toOffset offset` since UInt64 contains [weak type information](https://swift.org/documentation/api-design-guidelines/#weak-type-information), but besides that I think seek is fine.

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -21,12 +21,57 @@ extension FileHandle {
...
+    func readDataUp(toLength length: Int) throws -> Data {
+        if #available(iOS 13.0, tvOS 13.0, watchOS 6.0, OSX 15.0, *) {
+            return try __readDataUp(toLength: length)
+        } else {
+            return try ObjectiveC.unsafe { readData(ofLength: length) }
+        }
+    }
+
+    /// A method to write data to a file handle that is safe to call in Swift from any operation system version.
+    func write(data: Data) throws {
```
fdiaz: nit: This can probably be just `write(_ data: Data)`

Tests/CacheAdvanceTests/TestableMessage.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+struct TestableMessage: Codable, CustomStringConvertible, Equatable {
```
fdiaz: I think you want [ExpressibleByStringLiteral](https://developer.apple.com/documentation/swift/expressiblebystringliteral) here instead of `CustomStringConvertible`?

That way, you don't need to change these https://github.com/dfed/CacheAdvance/pull/19/commits/8697adf1142c8c0cab58002828478518dbdaea33#diff-57e5815eda868a2058fc13cd1e7dbb46R220

Scripts/build.swift
```diff
@@ -20,38 +20,132 @@ enum TaskError: Error {
...
+    case watchOS_5
+    case watchOS_6
+
+    var destination: String {
+        switch self {
+        case .iOS_12:
```
fdiaz: Just curious, why did you decide to use iPad instead of iPhones?

Scripts/build.swift
```diff
@@ -20,38 +20,132 @@ enum TaskError: Error {
...
+            return false
+
+        case .watchOS_5,
+             .watchOS_6:
+            // watchOS does not support unit testing (yet?).
+            return false
+        }
+    }
+
+    var enableCodeCoverage: Bool {
```
fdiaz: A comment here would be great re: why we're only gathering code coverage for tv OS

Scripts/build.swift
```diff
@@ -20,38 +20,132 @@ enum TaskError: Error {
...
+        xcodeBuildArguments.append("-derivedDataPath")
+        xcodeBuildArguments.append(".build/derivedData")
+    }
+    xcodeBuildArguments.append("build")
+    if platform.shouldTest {
+        xcodeBuildArguments.append("test")
+    }
+
+    try execute(commandPath: "/usr/bin/xcodebuild", arguments: xcodeBuildArguments)
+}
```
fdiaz: This is really clean 👍

Sources/CacheAdvance/FileHandleExtensions.swift
dfed: @fdiaz I agree! But iOS 13's API already took that name 😂

Sources/CacheAdvance/FileHandleExtensions.swift
dfed: Yes! But this is Apple's API signature and I needed a different signature here.

Scripts/build.swift
dfed: CI machines tend to have multiple OS installs for a particular iPad, but less so for particular iPhones.

Scripts/build.swift
dfed: I'll add something, though the reasoning is "there was no need to do more than this" 😅

dfed:
Merging! If y'all have ideas for better names on the FileHandle extensions, I'd love to hear it (and will address in a follow-up!).

Sources/SwiftTryCatch/SwiftTryCatch.h
bachand: Nice, I’m glad we’re attributing the original author.

PR #20, opened 2019-12-23
Do not build branch build on every push on PR branches

dfed:
We don't need to build both a `pr` build and a `push` build on every pull request. This PR makes it so we only build `push` builds for `master`. We'll still have PR builds on every PR, but we won't build every PR build twice.

See [Travis's docs](https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches) for more info on this change.

cc @bachand

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,7 @@ matrix:
     env: ACTION="pod-lint";SWIFT_VERSION="5.1"
   - osx_image: xcode11.2
     env: ACTION="carthage"
+
+branches:
+  only:
+  - master


bachand:
Nice. Is the PR build still governed by `.travis.yml`? I didn’t see any mention of it in the linked Travis docs. Haven’t used Travis much so this may be a n00b question.

dfed:
From what I've seen on Valet, `push` builds are made on any branch in the `only` list. PR builds are made whenever a PR targets a branch in the `only` list. This means that stacked PRs don't get builds automatically, but that's not the end of the world.

bachand:
Cool. Good to know.
Feels like there could be a more obvious setup. Seems odd that the default is for each PR to do two builds 🤔
Glad we’re simplifying.

PR #21, opened 2019-12-26
Fix macOS availability

dfed:
We were checking for `OSX 15.0`, when we meant to test for `OSX 10.15`.

While I was here, I updated `OSX` to `macOS`, which is synonymous and more up-to-date with Apple's marketing.


bachand:
A bummer it wasn’t caught in CI but makes sense why it wasn’t.

PR #23, opened 2019-12-30
Fill in the README

dfed:
This PR shouldn't be merged until #22 is merged (or if #22 is not merged I'll need to update the version numbers specified under `Installation`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,9 @@ If a `CacheAdvanceError.fileCorrupted` error is thrown, the cache file is corrup
 
 ## How it works
 
-CacheAdvance immediately persists each appended messages to disk using `FileHandle`s. More information to come!
+CacheAdvance immediately persists each appended messages to disk using `FileHandle`s. Messages are encoded using a `JSONEncoder`. Messages are written to disk as an encoded data blob that is prefixed with the length of the message. The length of a message is stored using a `UInt32` to ensure that the size of the data on disk that stores a message's length is consistent between devices.
+
+The first 64bytes of a CacheAdvance is reserved for storing metadata about the file. Any configuration data that must be static between cache opens should be stored in this header. It is also reasonable to store mutable information in the header, if doing so speeds up reads or writes to the file. The header format is managed by [FileHeader.swift](Sources/CacheAdvance/FileHeader.swift).
 
 ## Requirements
 
--- a/README.md
+++ b/README.md
@@ -1,2 +1,112 @@
 # CacheAdvance
+
+[![CI Status](https://travis-ci.com/dfed/CacheAdvance.svg?branch=master)](https://travis-ci.com/dfed/CacheAdvance)
+[![Swift Package Manager compatible](https://img.shields.io/badge/SPM-compatible-4BC51D.svg?style=flat)](https://github.com/apple/swift-package-manager)
+[![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
+[![Version](https://img.shields.io/cocoapods/v/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
+[![License](https://img.shields.io/cocoapods/l/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
+[![Platform](https://img.shields.io/cocoapods/p/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
+
 A cache that enables the performant persistence of individual messages to disk.
+
+## Usage
+
+### Basic Initialization
+
+```swift
+let myCache = try CacheAdvance<MyMessageType>(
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,112 @@
...
+
+### Error handling
+
+A CacheAdvance will never fatal error: only recoverable errors will be thrown. A CacheAdvance may throw a `CacheAdvanceError`, or errors related to reading or writing with `FileHandle`s.
+
+If a `CacheAdvanceError.fileCorrupted` error is thrown, the cache file is corrupt and should be deleted.
+
+## How it works
+
+CacheAdvance immediately persists each appended messages to disk using `FileHandle`s. More information to come!
```
dfed: Things we should talk about here:
* How messages are encoded
* File format

dfed: https://github.com/dfed/CacheAdvance/pull/23/commits/6447d5590ae153effd92a9ee5235b1a487f04410

README.md
```diff
@@ -1,2 +1,116 @@
 # CacheAdvance
+
+[![CI Status](https://travis-ci.com/dfed/CacheAdvance.svg?branch=master)](https://travis-ci.com/dfed/CacheAdvance)
+[![Swift Package Manager compatible](https://img.shields.io/badge/SPM-compatible-4BC51D.svg?style=flat)](https://github.com/apple/swift-package-manager)
+[![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
+[![Version](https://img.shields.io/cocoapods/v/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
```
bachand: The pod logo is orange. Is that bad?

README.md
```diff
@@ -1,2 +1,116 @@
...
+
+```swift
+try myCache.append(message: aMessageInstance)
+```
+
+By the time the above method exits, the message will have been persisted to disk. A CacheAdvance keeps no in-memory buffer. Appending a new message is cheap, as a CacheAdvance needs to encode and persist only the new message and associated metadata.
+
+A CacheAdvance instance that does not overwrite old messages will throw a `CacheAdvanceError.messageDataTooLarge` if appending a message would exceed the cache's `maximumBytes`. A CacheAdvance instance that does overwrite old messages will throw a `CacheAdvanceError.messageDataTooLarge` if the message would require more than `maximumBytes` to store even after evicting all older messages from the cache.
+
+To ensure that caches can be read from 32bit devices, messages should not be larger than 2GB in size.
```
bachand: Do we assert for this in code? Or maybe we should throw an exception?

Also are we sure this restriction is real? I thought a 32 bit device still has the concept of `UInt64` but that it was less efficient.

README.md
```diff
@@ -1,2 +1,116 @@
...
+* tvOS 12 or later.
+* watchOS 5 or later.
+* macOS 10.14 or later.
+* Swift 5.0 or later.
+
+## Installation
+
+### Swift Package Manager
+
+To install Relativity in your iOS project with [Swift Package Manager](https://github.com/apple/swift-package-manager), the following lines can be added to your `Package.swift` file:
```
bachand: Cache Advance not Relativity.

README.md
```diff
@@ -1,2 +1,116 @@
...
+
+Run `carthage` to build the framework and drag the built `CacheAdvance.framework` into your Xcode project.
+
+### Submodules
+
+To use git submodules, checkout the submodule with `git submodule add git@github.com:dfed/CacheAdvance.git`, drag CacheAdvance.xcodeproj to your project, and add Relativity as a build dependency.
+
+## Contributing
+
+I’m glad you’re interested in CacheAdvance, and I’d love to see where you take it. Please read the [contributing guidelines](Contributing.md) prior to submitting a Pull Request.
```
bachand: :)

README.md
dfed: I think it just means beta?

README.md
dfed: We don't assert or throw an exception. Starting with documentation. I can make a separate to enforce this.

32-bit devices do have the concept of `UInt64`, but `Int` will be an `Int32` on a 32-bit device, which means that when we cast the message length to `Int` in order to `readDataUp(toLength:)`, we'll overflow on anything larger than ~2GB.

README.md
dfed: 😂 I gotta get better at copy/paste. Thank you for catching!

README.md
dfed: Added a `throw`. See https://github.com/dfed/CacheAdvance/pull/22#discussion_r362278580.

That said, we may still have an issue on 32bit devices, since I have no idea what happens when someone tries to encode a Data blob longer than `Int32`. I'm hoping that trying to encode a message longer than `Int32` on a 32bit device `throws`, but I don't have a good way to check.

bachand: > I'm hoping that trying to encode a message longer than `Int32` on a 32bit device throws, but I don't have a good way to check.

I see what you’re saying. Hopefully it `throws`...

Issue #2, opened 2019-12-02
Get tests running on macOS in CI


dfed:
We currently do not run unit tests on our [macOS CI job](https://github.com/dfed/CacheAdvance/blob/35aa75a01cabfcf84f5d64f2c318337a294ebb02/.travis.yml#L14). This is because we're using Travis CI, and Travis CI does not yet have machines with Catalina (macOS 10.15) – it's latest machines run macOS 10.14. Our base SDK on macOS is macOS 10.15 due to SDK limitations, our tests fail on macOS 10.14.

There are three possible paths forward for fixing this:
1. Move to Circle CI. Circle CI has machines with macOS 10.15.
1. Lower our base SDK to macOS 10.14. We'll likely want to do this anyways.
1. Wait for Travis CI to start deploying macOS 10.15.

This issue tracks enabling the `test` parameter on our macOS CI job.

dfed:
I have a PR out to lower our base SDK to macOS 10.14, but since this project doesn't support being built with Xcode 10, we can't run tests on macOS 10.14.

Crossing my fingers that Travis CI gets macOS 10.15 machines soon.

PR #24, opened 2019-12-30
Run tests on macOS

dfed:
This resolves https://github.com/dfed/CacheAdvance/issues/2

Now that we have `if #available` checks to guard against using the new APIs on older operating systems, Travis' macOS machine will register as a `10.14` machine, and we won't crash.

We’re still not testing on `10.15`, but testing on multiple macOS versions when only recent Xcode versions are supported is an unsolved problem.

### DIFF
--- a/Scripts/build.swift
+++ b/Scripts/build.swift
@@ -75,13 +75,9 @@ enum Platform: String, CaseIterable, CustomStringConvertible {
         case .iOS_12,
              .iOS_13,
              .tvOS_12,
-             .tvOS_13:
+             .tvOS_13,
+             .macOS_10_15:
             return true
-
-        case .macOS_10_15:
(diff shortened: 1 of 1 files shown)


dfed:
This worked. I'm going to merge!

bachand:
> I have a PR out to lower our base SDK to macOS 10.14, but since this project doesn't support being built with Xcode 10, we can't run tests on macOS 10.14.

How did we resolve the Xcode 10 issue?

dfed:
We didn’t. We’re building on a macOS 10.14 with Xcode 11, and then running tests on that macOS 10.14 machine.

Travis’s setup is absurd. But since our code has the right availability code, and is running on the macOS 10.14 SDK, we don’t crash.

What I’d missed was that the Xcode we needed was already running on a supported version of macOS.

PR #25, opened 2020-01-14
Provide development instructions and ignore Xcode projects

bachand:
I ran `swift package generate-xcodeproj` from the root of the repo. That generated `CacheAdvance.xcodeproj`, which I can use to build and run this project.

Since we can generate this project from the package, it seems like we don't need to check it in.

While I was at it, I added a section on Developing to the README, and described how to get a working Xcode project.

Please let me know if I'm doing anything in a non-standard way. I was interested in poking around the code in my IDE and this is how I would normally get started.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,5 +21,5 @@ DerivedData
 generated/
 *coverage.txt
 
-# We can generate the project from the package. No need to check it in.
+# We can generate Xcode projects from the Swift packages.
 *.xcodeproj/
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ DerivedData
 .swiftpm/
 generated/
 *coverage.txt
+
+# We can generate the project from the package. No need to check it in.
+CacheAdvance.xcodeproj/
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,4 @@ generated/
 *coverage.txt
 
 # We can generate the project from the package. No need to check it in.
-CacheAdvance.xcodeproj/
+*.xcodeproj/
--- a/README.md
+++ b/README.md
@@ -111,6 +111,10 @@ I’m glad you’re interested in CacheAdvance, and I’d love to see where you 
 
 Thanks, and happy caching!
 
+## Developing
+
+Run `swift package generate-xcodeproj` from the root of the repository to generate an Xcode project.
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -111,6 +111,10 @@ I’m glad you’re interested in CacheAdvance, and I’d love to see where you
 
 Thanks, and happy caching!
 
+## Developing
+
+Run `swift package generate-xcodeproj` from the root of the repository to generate an Xcode project.
```
dfed: I wouldn’t recommend using the manually generated project if we don’t have to. Opening the `Package.swift` is magical, and it live-reloads. Since our minimum Xcode support is Xcode 11, I’m thinking we should suggest the following:
```suggestion
Double-click on `Package.swift` in the root of the repository to open the project in Xcode.
```

.gitignore
```diff
@@ -20,3 +20,6 @@ DerivedData
 .swiftpm/
 generated/
 *coverage.txt
+
+# We can generate the project from the package. No need to check it in.
+CacheAdvance.xcodeproj/
```
dfed: Smart thing to ignore. We don’t need any xcode projects: should we ignore any project, no matter where it is located?
```suggestion
*.xcodeproj/
```

README.md
bachand: TIL. I didn't know you could do that.

.gitignore
bachand: I was on the fence about this. Let's do what you suggested.
I'll simplify the comment a bit too.

README.md
dfed: Now you see why I’m convinced Swift Package Manager is the future 🙂

bachand: Indeed

bachand:
@dfed do you have any idea why the code coverage bot reports a decrease in code coverage?

dfed:
@bachand my theory is that because I’m uploading coverage info from two jobs to get the complete coverage, maybe one of the jobs got far enough to upload one but didn’t upload the other before being cancelled?

Your latest commit seems to have resolved the issue, so hard for me to be sure.

bachand:
The would make some sense.

README.md
fdiaz: This is cool! I didn't know about that either!

PR #27, opened 2020-01-15
Use one machine to get code coverage for two jobs

dfed:
In this PR I try to use a single travis configuration to get code coverage for multiple builds. This should hopefully improve the reliability of the code coverage comment, preventing issues like https://github.com/dfed/CacheAdvance/pull/25#issuecomment-574007303.

Since we can now upload test coverage for multiple jobs on the same machine, it seemed reasonable to enable test coverage on more jobs.

### DIFF
--- a/Scripts/build.swift
+++ b/Scripts/build.swift
@@ -86,22 +86,6 @@ enum Platform: String, CaseIterable, CustomStringConvertible {
         }
     }
 
-    var enableCodeCoverage: Bool {
-        switch self {
-        case .tvOS_12,
-             .tvOS_13:
-            // We currently only have a after_success job to upload code coverage for our tvOS targets.
-            return true
-
-        case .iOS_12,
-             .iOS_13,
-             .macOS_10_15,
-             .watchOS_5,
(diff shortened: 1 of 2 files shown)


Scripts/build.swift
```diff
@@ -133,11 +121,11 @@ for rawPlatform in rawPlatforms {
         xcodeBuildArguments.append("-destination")
         xcodeBuildArguments.append(platform.destination)
     }
-    if platform.enableCodeCoverage {
+    if platform.shouldTest {
         xcodeBuildArguments.append("-enableCodeCoverage")
         xcodeBuildArguments.append("YES")
         xcodeBuildArguments.append("-derivedDataPath")
-        xcodeBuildArguments.append(".build/derivedData")
+        xcodeBuildArguments.append(platform.derivedDataPath)
```
bachand: I wonder if we should always set the custom derived data path even if we aren't testing, just to make things simpler?

dfed: Can't hurt to always set the derived data. Will make that change.

PR #29, opened 2020-01-21
Simplify and refactor `synchronizeHeaderData()` in preparation for `isCompatible(...)` method

bachand:
I'm planning to add an `isCompatible(...)` public method to the `CacheAdvance` type. See the code at the bottom of this PR message. I almost finished adding it in this PR, but I'm trying to think about how to avoid duplicating too much logic between tests and I ran out of time now. Many of the same tests that we added for the `canOpenFile(...)` method (see https://github.com/dfed/CacheAdvance/pull/28) would apply here as well. Any ideas would be appreciated. Also I welcome any feedback on this method before I put up that PR.

This PR further refactors `CacheHeaderHandle` in preparation for adding the `isCompatible(...)` method. I've extracted the code that reads the data for the header into its own method. Previously we had a separate code path for resetting the file if it is empty. I added at test for that case and then removed that code path in favor of the code we already had, which would do what we wanted.

----

```swift
public static func isCompatible(
    withFile file: URL,
    maximumBytes: Bytes,
    shouldOverwriteOldMessages: Bool)
    throws -> Bool
{
    let cacheHeaderHandle = try CacheHeaderHandle(
        forReadingFrom: file,
        maximumBytes: maximumBytes,
        overwritesOldMessages: shouldOverwriteOldMessages)

    let headerData = try cacheHeaderHandle.readHeaderData()

    guard let fileHeader = FileHeader(from: headerData) else {
        // We can't read the header data. This is definitely not compatible.
        return false
    }

    guard cacheHeaderHandle.canOpenFile(with: fileHeader) else {
        // The header is incompatible.
        return false
    }

    return true
}
```

### DIFF
--- a/Sources/CacheAdvance/CacheHeaderHandle.swift
+++ b/Sources/CacheAdvance/CacheHeaderHandle.swift
@@ -114,26 +114,20 @@ final class CacheHeaderHandle {
     func synchronizeHeaderData() throws {
         let headerData = try readHeaderData()
 
-        if headerData.isEmpty {
-            // There is no header. Write one to disk.
-            try writeHeaderData()
-
-        } else {
-            guard let fileHeader = FileHeader(from: headerData) else {
-                // We can't read the header data. Let's start over.
-                try resetFile()
-                return
-            }
-
-            guard canOpenFile(with: fileHeader) else {
-                // The header is invalid. Nuke it.
-                try resetFile()
-                return
-            }
-
-            self.offsetInFileOfOldestMessage = fileHeader.offsetInFileOfOldestMessage
-            self.offsetInFileAtEndOfNewestMessage = fileHeader.offsetInFileAtEndOfNewestMessage
+        guard let fileHeader = FileHeader(from: headerData) else {
+            // We can't read the header data. Let's start over.
+            try resetFile()
+            return
         }
+
+        guard canOpenFile(with: fileHeader) else {
+            // The header is invalid. Nuke it.
+            try resetFile()
+            return
+        }
+
+        self.offsetInFileOfOldestMessage = fileHeader.offsetInFileOfOldestMessage
+        self.offsetInFileAtEndOfNewestMessage = fileHeader.offsetInFileAtEndOfNewestMessage
     }
 
     // MARK: Private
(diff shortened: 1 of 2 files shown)


Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -101,32 +112,22 @@ final class CacheHeaderHandle {
...
     /// If the header data persisted to the file is not consistent with expectations, the file will be deleted.
     func synchronizeHeaderData() throws {
-        // Start at the beginning of the file.
-        try handle.seek(to: 0)
+        let headerData = try readHeaderData()
 
-        // Read the entire header.
-        let headerData = try handle.readDataUp(toLength: Int(FileHeader.expectedEndOfHeaderInFile))
-
-        if headerData.isEmpty {
```
bachand: We don't need this code path. I added a test for it, then I removed it. We will fail to create a header if the data is empty, and then we'll invoke `resetFile`. The only impact of this change is that we will invoke `handle.truncate(at: 0)` when the file is already empty, but I don't think that's something we need to worry about. The test proves that it works.

bachand:
This PR is ready for review.

bachand:
@dfed I guess the code coverage went down because I changed the number of lines?

dfed:
> I'm trying to think about how to avoid duplicating too much logic between tests.... Many of the same tests that we added for the `canOpenFile(...)` method (see https://github.com/dfed/CacheAdvance/pull/28) would apply here as well

I'd be down for moving some of the tests currently on `canOpenFile(with:` to the `isCompatible`. Since we have code coverage tracking, I'm less convinced we need full tests for individual methods, as much as we should be testing how the whole system works.

dfed:
> I guess the code coverage went down because I changed the number of lines?

Yeah that sounds right. That's an unfortunate side-effect of not having 100% coverage I guess 🤔

bachand:
> I'm less convinced we need full tests for individual methods, as much as we should be testing how the whole system works.

I agree with this ^

> I'd be down for moving some of the tests currently on `canOpenFile(with:` to the `isCompatible`. 

I'm leaning towards leaving the detailed tests of all of the mismatch combinations on `canOpenFile(...)` since they serve as a nice form of documentation. Since the method will be invoked in multiple places (`synchronizeHeaderData()` and `isCompatible(...)`) it feels like the detailed tests best belong on the method itself, since it's not clear why they would be on `synchronizeHeaderData()` vs. `isCompatible(...)`. I am thinking of maybe adding a single way for `canOpenFile(...)` to fail in my `isCompatible(...)` tests (vs. all possible mismatches), building off of what you said about not having full tests for each method.

Anyway we can discuss more in the next PR, and maybe my thoughts will evolve as I write it.

PR #30, opened 2020-01-22
Add API to determine if a cache isEmpty()

dfed:
I'm working on integrating CacheAdvance, and realized we're missing the ability to check if the cache has any items in it.

### DIFF
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -24,6 +24,47 @@ final class CacheAdvanceTests: XCTestCase {
 
     // MARK: Behavior Tests
 
+    func test_isEmpty_returnsTrueWhenCacheIsEmpty() throws {
+        let cache = try createCache()
+
+        XCTAssertTrue(try cache.isEmpty())
+    }
+
+    func test_isEmpty_returnsFalseWhenCacheHasASingleMessage() throws {
+        let message: TestableMessage = "This is a test"
+        let cache = try createCache(messages: [message])
+        try cache.append(message: message)
+
+        XCTAssertFalse(try cache.isEmpty())
+    }
+
+    func test_isEmpty_returnsFalseWhenOpenedOnCacheThatHasASingleMessage() throws {
+        let message: TestableMessage = "This is a test"
+        let cache = try createCache()
+        try cache.append(message: message)
+
+        let sameCache = try createCache(zeroOutExistingFile: false)
+
+        XCTAssertFalse(try sameCache.isEmpty())
+    }
+
+    func test_isEmpty_returnsFalseWhenCacheThatDoesNotOverwriteIsFull() throws {
+        let message: TestableMessage = "This is a test"
+        let cache = try createCache(messages: [message])
+        try cache.append(message: message)
+
+        XCTAssertFalse(try cache.isEmpty())
+    }
+
+    func test_isEmpty_returnsFalseWhenCacheThatOverwritesIsFull() throws {
+        let cache = try createCache(shouldOverwriteOldMessages: true)
+        for message in Self.lorumIpsumMessages {
+            try cache.append(message: message)
+        }
+
+        XCTAssertFalse(try cache.isEmpty())
+    }
+
     func test_messages_canReadEmptyCacheThatDoesNotOverwriteOldestMessages() throws {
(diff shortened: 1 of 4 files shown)


dfed:
Reverting the version bump. Thinking I'll want to get more things into this release.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -24,6 +24,47 @@ final class CacheAdvanceTests: XCTestCase {
...
+        let message: TestableMessage = "This is a test"
+        let cache = try createCache()
+        try cache.append(message: message)
+
+        let sameCache = try createCache(zeroOutExistingFile: false)
+
+        XCTAssertFalse(try sameCache.isEmpty())
+    }
+
+    func test_isEmpty_returnsFalseWhenCacheThatDoesNotOverwriteIsFull() throws {
```
bachand: I don’t think this cache is full, right?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -24,6 +24,47 @@ final class CacheAdvanceTests: XCTestCase {
...
+        let cache = try createCache(messages: [message])
+        try cache.append(message: message)
+
+        XCTAssertFalse(try cache.isEmpty())
+    }
+
+    func test_isEmpty_returnsFalseWhenCacheThatOverwritesIsFull() throws {
+        let cache = try createCache(shouldOverwriteOldMessages: true)
+        for message in Self.lorumIpsumMessages {
+            try cache.append(message: message)
```
bachand: I forget how exactly how these tests work (on mobile). Does iterating through all of the messages make the cache full?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -24,6 +24,47 @@ final class CacheAdvanceTests: XCTestCase {
...
+
+    func test_isEmpty_returnsFalseWhenCacheHasASingleMessage() throws {
+        let message: TestableMessage = "This is a test"
+        let cache = try createCache(messages: [message])
+        try cache.append(message: message)
+
+        XCTAssertFalse(try cache.isEmpty())
+    }
+
+    func test_isEmpty_returnsFalseWhenOpenedOnCacheThatHasASingleMessage() throws {
```
bachand: 👌

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -24,6 +24,47 @@ final class CacheAdvanceTests: XCTestCase {
...
+        try cache.append(message: message)
+
+        let sameCache = try createCache(zeroOutExistingFile: false)
+
+        XCTAssertFalse(try sameCache.isEmpty())
+    }
+
+    func test_isEmpty_returnsFalseWhenCacheThatDoesNotOverwriteIsFull() throws {
+        let message: TestableMessage = "This is a test"
+        let cache = try createCache(messages: [message])
```
bachand: It could be nice to pass `shouldOverwriteOldMessages: false` (which I assume is the default) to make the intent more clear.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand: This seems to be true.

dfed: `createCache` using the default `messages` parameter sets the size to the size required to fit `Self.lorumIpsumMessages`

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: `try createCache(messages: [message])` creates a cache that is of the exact size to store `[message]`. So this cache is full.

I can do a separate PR to make these parameters a bit more intuitive. It confused me while I was writing this too 😅

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: it is indeed the default. I'll think I'll remove the default parameter in a separate PR. The defaults and parameter names as-is are... a bit less intuitive than I originally imagined.

PR #31, opened 2020-01-22
Clearly separate cache full error from message too large error

dfed:
When trying to adopt this project, I realized that we had conflated two similar errors that required differentiated error handling. `messageDataTooLarge` covered both "this message could never fit in this cache" and "the remaining space left in the cache can't fit this message".

In this PR, I separate this case into two separate errors.

This is a source-breaking change. Our next version will need to be 0.2.0 if we merge this PR as-is.

### DIFF
--- a/Sources/CacheAdvance/EncodableMessage.swift
+++ b/Sources/CacheAdvance/EncodableMessage.swift
@@ -42,7 +42,7 @@ struct EncodableMessage<T: Codable> {
         let messageData = try encoder.encode(message)
         guard messageData.count < MessageSpan.max else {
             // We can't encode the length this message in a MessageSpan.
-            throw CacheAdvanceError.messageDataTooLarge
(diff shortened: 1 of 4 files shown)


Sources/CacheAdvance/EncodableMessage.swift
```diff
@@ -42,7 +42,7 @@ struct EncodableMessage<T: Codable> {
         let messageData = try encoder.encode(message)
         guard messageData.count < MessageSpan.max else {
             // We can't encode the length this message in a MessageSpan.
-            throw CacheAdvanceError.messageDataTooLarge
+            throw CacheAdvanceError.messageLargerThanCacheCapacity
```
bachand: So `MessageSpan.max` Is the max value that can be stored in a `UInt32` IIUC.
I guess we could only hit this on a 64-bit device?

dfed: Correct

PR #37, opened 2020-01-24
Enable reading messages from a cache with unexpected header values

### LINKED ISSUE #35: Enable reading messages from a cache with unexpected header values
**Problem statement**
Today, we delete an opened cache file if the `maximumBytes` or `overwritesOldMessages` is not the value we expect when we open the file. This choice forces applications to choose between never changing the values of `maximumBytes` and `overwritesOldMessages`, or changing those values and losing their data. Providing fine-tuning of cache sizes and configurations without preventing reading caches that are already on-disk feels like a win.

**Goals**
* Enable applications to change `maximumBytes` or `overwritesOldMessages` without losing old cached data.

**Non-Goals**
* Enable writing to a cache when `maximumBytes` or `overwritesOldMessages`  has changed.

**Design**
Rather than resetting a file if `maximumBytes` or `overwritesOldMessages` have changed, we can store that these values have changed in memory. When an application attempts to write to a cache where the `maximumBytes` or `overwritesOldMessages` values are not consistent with the values provided in the `init` method, the `CacheAdvance` can throw a `CacheAdvanceError.headerOutOfSync` error. Reading messages, however, could still return the expected values.

dfed:
Resolves #35.

### DIFF
--- a/Sources/CacheAdvance/CacheAdvance.swift
+++ b/Sources/CacheAdvance/CacheAdvance.swift
@@ -64,6 +64,9 @@ public final class CacheAdvance<T: Codable> {
     /// - Parameter message: A message to write to disk. Must be smaller than both `maximumBytes - FileHeader.expectedEndOfHeaderInFile` and `MessageSpan.max`.
     public func append(message: T) throws {
         try setUpFileHandlesIfNecessary()
+        guard try header.canOpenFile() else {
+            throw CacheAdvanceError.incompatibleHeader
+        }
         guard try header.canWriteToFile() else {
             throw CacheAdvanceError.fileNotWritable
         }
@@ -130,6 +133,9 @@ public final class CacheAdvance<T: Codable> {
     /// - Returns: `true` when there are no messages written to the file.
     public func isEmpty() throws -> Bool {
         try setUpFileHandlesIfNecessary()
+        guard try header.canOpenFile() else {
+            throw CacheAdvanceError.incompatibleHeader
+        }
 
         return header.offsetInFileAtEndOfNewestMessage == FileHeader.expectedEndOfHeaderInFile
     }
@@ -137,6 +143,9 @@ public final class CacheAdvance<T: Codable> {
     /// Fetches all messages from the cache.
     public func messages() throws -> [T] {
         try setUpFileHandlesIfNecessary()
+        guard try header.canOpenFile() else {
+            throw CacheAdvanceError.incompatibleHeader
+        }
 
         var messages = [T]()
         while let encodedMessage = try reader.nextEncodedMessage() {
--- a/Sources/CacheAdvance/CacheAdvance.swift
+++ b/Sources/CacheAdvance/CacheAdvance.swift
@@ -39,10 +39,9 @@ public final class CacheAdvance<T: Codable> {
         throws
     {
         self.fileURL = fileURL
-        self.shouldOverwriteOldMessages = shouldOverwriteOldMessages
 
         writer = try FileHandle(forWritingTo: fileURL)
-        reader = try CacheReader(forReadingFrom: fileURL, overwriteOldMessages: shouldOverwriteOldMessages)
+        reader = try CacheReader(forReadingFrom: fileURL)
         header = try CacheHeaderHandle(forReadingFrom: fileURL, maximumBytes: maximumBytes, overwritesOldMessages: shouldOverwriteOldMessages)
     }
 
@@ -83,7 +82,7 @@ public final class CacheAdvance<T: Codable> {
         }
 
         let cacheHasSpaceForNewMessageBeforeEndOfFile = writer.offsetInFile + bytesNeededToStoreMessage <= header.maximumBytes
-        if shouldOverwriteOldMessages {
+        if header.overwritesOldMessages {
             if !cacheHasSpaceForNewMessageBeforeEndOfFile {
                 // This message can't be written without exceeding our maximum file length.
                 // We'll need to start writing the file from the beginning of the file.
@@ -208,7 +207,6 @@ public final class CacheAdvance<T: Codable> {
     private let header: CacheHeaderHandle
 
     private var hasSetUpFileHandles = false
-    private let shouldOverwriteOldMessages: Bool
 
     private let decoder = JSONDecoder()
     private let encoder = JSONEncoder()
--- a/Sources/CacheAdvance/CacheAdvance.swift
+++ b/Sources/CacheAdvance/CacheAdvance.swift
@@ -30,8 +30,8 @@ public final class CacheAdvance<T: Codable> {
     ///   - maximumBytes: The maximum size of the cache, in bytes. Logs larger than this size will fail to append to the store.
     ///   - shouldOverwriteOldMessages: When `true`, once the on-disk store exceeds maximumBytes, new entries will replace the oldest entry.
     ///
-    /// - Warning: `maximumBytes` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will lead to data loss.
-    /// - Warning: `shouldOverwriteOldMessages` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will lead to data loss.
+    /// - Warning: `maximumBytes` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will prevent appending new messages to this cache.
+    /// - Warning: `shouldOverwriteOldMessages` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will prevent appending new messages to this cache.
     public init(
         fileURL: URL,
         maximumBytes: Bytes,
@@ -39,7 +39,6 @@ public final class CacheAdvance<T: Codable> {
         throws
     {
         self.fileURL = fileURL
-        self.maximumBytes = maximumBytes
         self.shouldOverwriteOldMessages = shouldOverwriteOldMessages
 
         writer = try FileHandle(forWritingTo: fileURL)
@@ -55,17 +54,27 @@ public final class CacheAdvance<T: Codable> {
 
     public let fileURL: URL
 
+    /// Checks if the all the header metadata provided at initialization matches the persisted header. If not, the cache is not writable.
+    /// - Returns: `true` if the cache is writable.
+    public func isWritable() throws -> Bool {
+        try setUpFileHandlesIfNecessary()
+        return try header.canWriteToFile()
+    }
+
     /// Appends a message to the cache.
     /// - Parameter message: A message to write to disk. Must be smaller than both `maximumBytes - FileHeader.expectedEndOfHeaderInFile` and `MessageSpan.max`.
     public func append(message: T) throws {
         try setUpFileHandlesIfNecessary()
+        guard try header.canWriteToFile() else {
+            throw CacheAdvanceError.fileNotWritable
+        }
 
         let encodableMessage = EncodableMessage(message: message, encoder: encoder)
         let messageData = try encodableMessage.encodedData()
         let bytesNeededToStoreMessage = Bytes(messageData.count)
 
         guard
-            bytesNeededToStoreMessage <= maximumBytes - FileHeader.expectedEndOfHeaderInFile // Make sure we have room in this file for this message.
+            bytesNeededToStoreMessage <= header.maximumBytes - FileHeader.expectedEndOfHeaderInFile // Make sure we have room in this file for this message.
                 && bytesNeededToStoreMessage < Int32.max // Make sure we can read this message back out with Int on a 32-bit device.
             else
         {
@@ -73,7 +82,7 @@ public final class CacheAdvance<T: Codable> {
             throw CacheAdvanceError.messageLargerThanCacheCapacity
         }
 
-        let cacheHasSpaceForNewMessageBeforeEndOfFile = writer.offsetInFile + bytesNeededToStoreMessage <= maximumBytes
+        let cacheHasSpaceForNewMessageBeforeEndOfFile = writer.offsetInFile + bytesNeededToStoreMessage <= header.maximumBytes
         if shouldOverwriteOldMessages {
             if !cacheHasSpaceForNewMessageBeforeEndOfFile {
                 // This message can't be written without exceeding our maximum file length.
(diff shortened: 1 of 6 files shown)


Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -49,20 +49,10 @@ final class CacheHeaderHandle {
...
     // MARK: Internal
 
+    let maximumBytes: Bytes
     private(set) var offsetInFileOfOldestMessage: UInt64
     private(set) var offsetInFileAtEndOfNewestMessage: UInt64
 
-    /// Attempts to read the header data of the file.
-    ///
-    /// - Returns: As much of the header data as exists. The returned data may not form a complete header.
-    func readHeaderData() throws -> Data {
```
dfed: This method was moved down under `Private` with no changes.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -487,8 +595,27 @@ final class CacheAdvanceTests: XCTestCase {
         }
     }
 
+    private func createHeaderHandle(
```
bachand: Could we consolidate with https://github.com/dfed/CacheAdvance/blob/bfe57b1649a0e3ed2303b8b399cc83206769049b/Tests/CacheAdvanceTests/CacheHeaderHandleTests.swift#L212-L224?

bachand: In some helper file

Tests/CacheAdvanceTests/CacheHeaderHandleTests.swift
```diff
@@ -49,7 +49,7 @@ final class CacheHeaderHandleTests: XCTestCase {
         XCTAssertEqual(headerHandle2.offsetInFileOfOldestMessage, 2000)
     }
 
-    func test_synchronizeHeaderData_returnsDefaultVersionWhenUnexpectedVersionIsOnDisk() throws {
+    func test_synchronizeHeaderData_throwsFileCorruptedWhenDefaultVersionWhenUnexpectedVersionIsOnDisk() throws {
```
bachand: Two "when"s in test name

bachand: I think I mentioned this elsewhere but it could be nice for the error that is thrown to be more specific. Is the file really corrupted?

Tests/CacheAdvanceTests/CacheHeaderHandleTests.swift
```diff
@@ -79,13 +78,10 @@ final class CacheHeaderHandleTests: XCTestCase {
         XCTAssertNotEqual(headerHandle1.offsetInFileOfOldestMessage, defaultOffsetInFileOfOldestMessage)
 
         let headerHandle2 = try createHeaderHandle(maximumBytes: 10000)
-        try headerHandle2.synchronizeHeaderData()
-
-        XCTAssertEqual(headerHandle2.offsetInFileAtEndOfNewestMessage, defaultOffsetInFileAtEndOfNewestMessage)
```
bachand: It could be nice to have some more guarantees around what should happen here.
Beyond not throwing I think it would be nice to know if the header gets rewritten or not.

Tests/CacheAdvanceTests/CacheHeaderHandleTests.swift
```diff
@@ -101,84 +97,114 @@ final class CacheHeaderHandleTests: XCTestCase {
...
+
+        let sut = try createHeaderHandle(
+            maximumBytes: 1000,
+            overwritesOldMessages: true,
+            version: 2)
+
+        XCTAssertFalse(try sut.canWriteToFile())
+    }
+
+    func test_canWriteToFile_emptyFile_returnsFalse() throws {
```
bachand: Can we add a test for when `canWriteToFile()` throws?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -55,26 +53,36 @@ public final class CacheAdvance<T: Codable> {
...
+            throw CacheAdvanceError.fileNotWritable
+        }
 
         let encodableMessage = EncodableMessage(message: message, encoder: encoder)
         let messageData = try encodableMessage.encodedData()
         let bytesNeededToStoreMessage = Bytes(messageData.count)
 
         guard
-            bytesNeededToStoreMessage <= maximumBytes - FileHeader.expectedEndOfHeaderInFile // Make sure we have room in this file for this message.
+            bytesNeededToStoreMessage <= header.maximumBytes - FileHeader.expectedEndOfHeaderInFile // Make sure we have room in this file for this message.
```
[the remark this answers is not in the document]
dfed: Good point. I was also thinking of renaming `CacheHeaderHandle` -> `CacheHeader`, since it's more than just a thin wrapper on a handle now. Will tackle both of these thoughts at the same time.

Sources/CacheAdvance/CacheAdvanceError.swift
```diff
@@ -20,7 +20,9 @@ public enum CacheAdvanceError: Error, Equatable {
     case messageLargerThanCacheCapacity
     /// Thrown when a message being appended to a cache that does not overwrite old messages is too large to store in the remaining space.
     case messageLargerThanRemainingCacheSize
+    /// Thrown when the cache file's persisted static header data is inconsistent with the metadata with which the cache was initialized.
+    case fileNotWritable
     /// Thrown when the cache file is of an unexpected format.
-    /// A corrupted file should be deleted. Corruption can occur if an application crashes while writing to the file.
+    /// A corrupted file should be deleted. Corruption can occur if an application crashes while writing to the file, or if a cache was persisted with an incompatible header.
```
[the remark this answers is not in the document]
dfed: > Since writes are atomic I thought we could only lose data in some cases

I believe it is possible for the application to crash in the middle of writing to a file handle.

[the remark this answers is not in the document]
dfed: yeah I think you're right. Updating.

Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -83,60 +74,55 @@ final class CacheHeaderHandle {
...
             return
         }
 
-        guard canOpenFile(with: fileHeader) else {
-            // The header is invalid. Nuke it.
-            try resetFile()
-            return
+        guard let fileHeader = FileHeader(from: headerData) else {
+            // We can't read the header data.
+            throw CacheAdvanceError.fileCorrupted
```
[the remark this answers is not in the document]
dfed: We'd recover by deleting the cache. I'm trying to move towards a world where we no longer delete data on disk, and instead report what went wrong and allow our consumers to handle as they please.

Now that I'm trying to integrate CacheAdvance into our app, I'm realizing that auto-delete isn't really acceptable, even if we're super clear when it would happen.

Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -83,60 +74,55 @@ final class CacheHeaderHandle {
...
     /// If the header data persisted to the file is not consistent with expectations, the file will be deleted.
     func synchronizeHeaderData() throws {
         let headerData = try readHeaderData()
 
-        guard let fileHeader = FileHeader(from: headerData) else {
-            // We can't read the header data. Let's start over.
-            try resetFile()
+        guard !headerData.isEmpty else {
+            // The file is empty. Write a header to disk.
+            try writeHeaderData()
```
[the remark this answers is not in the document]
dfed: yeah I like that

Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -155,13 +177,42 @@ final class CacheHeaderHandle {
...
-        try handle.truncate(at: 0)
-
-        // Now that we've started from scratch, write a new header.
-        try writeHeaderData()
+    private func memoizedMetadata() throws -> Metadata {
+        if let persistedMetadata = self.persistedMetadata {
+            return persistedMetadata
+        } else {
+            return try Metadata(headerData: try readHeaderData())
+        }
```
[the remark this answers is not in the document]
dfed: Since we're only memoizing `Metadata` that is static, we're fine.

Of course, another process could smash what is on disk and we'd get out of sync. But... I think that'd be user error.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -79,6 +79,45 @@ final class CacheAdvanceTests: XCTestCase {
...
 
+    func test_isWritable_returnsTrueWhenStaticHeaderMetadataMatches() throws {
+        let originalCache = try createCache(overwritesOldMessages: false)
+        XCTAssertTrue(try originalCache.isWritable())
+
+        let sut = try createCache(overwritesOldMessages: false, zeroOutExistingFile: false)
+        XCTAssertTrue(try sut.isWritable())
+    }
+
+    func test_isWritable_throwsFileCorruptedWhenHeaderVersionDoesNotMatch() throws {
```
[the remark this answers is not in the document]
dfed: `isWritable` needs to throw because we need to call `try setUpFileHandlesIfNecessary()` before reading the handle.

But I think you're right that we shouldn't be throwing `fileCorrupted` here. We should be returning `false` in this case.

Re whether we can swallow errors and not `throws` in this method... I could be convinced of that. The issue is if a client wants to know the difference between "version mismatch" and "the disk isn't readable", we'd be swallowing that.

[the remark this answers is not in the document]
dfed: Looks like we `throw` in `synchronizeHeaderData()`. I need to think about how to unwind this 🤔

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: The signatures are a bit different, since we're doing slightly different things. Thinking there's value in keeping this separate for now.

Tests/CacheAdvanceTests/CacheHeaderHandleTests.swift
dfed: yeah this shouldn't `throw`. Fixing.

Tests/CacheAdvanceTests/CacheHeaderHandleTests.swift
dfed: yeah I can ensure the file doesn't change. good call

Tests/CacheAdvanceTests/CacheHeaderHandleTests.swift
dfed: when I next push, this will only throw on a `FileHandle` error, which is not something I can simulate.

Sources/CacheAdvance/CacheAdvanceError.swift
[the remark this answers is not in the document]
bachand: interesting.

Sources/CacheAdvance/CacheHeaderHandle.swift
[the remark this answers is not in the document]
bachand: > Now that I'm trying to integrate CacheAdvance into our app, I'm realizing that auto-delete isn't really acceptable, even if we're super clear when it would happen.

Yeah I tend to agree.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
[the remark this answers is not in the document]
bachand: > Re whether we can swallow errors and not `throws` in this method... I could be convinced of that. The issue is if a client wants to know the difference between "version mismatch" and "the disk isn't readable", we'd be swallowing that.

Yeah I'm torn on stuff like this. I do err on the side of more information. Also wondering if there are ways to throw less. But it's hard to do both 😓

Sources/CacheAdvance/CacheAdvanceError.swift
```diff
@@ -20,9 +20,11 @@ public enum CacheAdvanceError: Error, Equatable {
     case messageLargerThanCacheCapacity
     /// Thrown when a message being appended to a cache that does not overwrite old messages is too large to store in the remaining space.
     case messageLargerThanRemainingCacheSize
+    /// Thrown when a cache's persisted header is incompatible with the current implementation.
+    case incompatibleHeader
     /// Thrown when the cache file's persisted static header data is inconsistent with the metadata with which the cache was initialized.
     case fileNotWritable
     /// Thrown when the cache file is of an unexpected format.
-    /// A corrupted file should be deleted. Corruption can occur if an application crashes while writing to the file, or if a cache was persisted with an incompatible header.
+    /// A corrupted file should be deleted. Corruption can occur if an application crashes while writing to the file.
```
bachand: 👍

Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -95,8 +95,8 @@ final class CacheHeaderHandle {
 
         guard !headerData.isEmpty else {
             // The file is empty. Write a header to disk.
-            try writeHeaderData()
-            persistedMetadata = Metadata(fileHeader: currentHeader)
+            let writtenHeader = try writeHeaderData()
+            persistedMetadata = Metadata(fileHeader: writtenHeader)
```
bachand: 💥

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -94,9 +108,7 @@ final class CacheAdvanceTests: XCTestCase {
         try originalHeader.synchronizeHeaderData()
 
         let sut = try createCache(overwritesOldMessages: false, zeroOutExistingFile: false)
```
bachand: We need to update the test name to reflect the new code.

Sources/CacheAdvance/CacheAdvance.swift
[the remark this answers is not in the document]
bachand: Yeah I like the idea of hiding the "handle" concept as much as possible. Hard to think about.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: Good catch! Fixed in 073f1c6

PR #38, opened 2020-01-24
Expose CacheReader to enable a more-explict read-only mode

dfed:
Built on top of #37. In this PR I expose the `CacheReader` publicly, in order to enable clients to open caches that were previously written to in an explicitly read-only mode.

I'm very much open to API and naming feedback.

I'm also open to better ways to handle the fact that the `CacheReader` now has the ability to read from a `CacheHeaderHandle`, but that it doesn't need to if the `CacheAdvance` has already set up the `CacheReader`. There's some weird interplay here that isn't the most obvious, but at a minimum it's all well-tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,13 @@ let cachedMessages = try myCache.messages()
 
 This method reads all cached messages from disk into memory.
 
+If you want to read messages that were previously cached and do not need to write new messages, you can utilize a `CacheReader` to read all cached messages into memory:
+
+```swift
+let myReader = try CacheReader(fileURL: FileManager.default.temporaryDirectory.appendingPathComponent("MyCache"))
+let cachedMessages = try myReader.messages()
+```
+
 ### Thread safety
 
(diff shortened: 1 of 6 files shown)


bachand:
I'm trying to understand the use case for this, since adding this will add complexity.

I could see having a `CacheAdvanceWriter` that _only_ writes and a `CacheAdvanceReader` that _only_ reads. But as we discussed in person, it's a bit scary to have two objects that are touching the same file on disk. Though since we are offloading thread safety to the consumer, maybe that's OK? I could be convinced that we have something that only reads and something that only writes.

If we exist in the current world where `CacheAdvance` can always read and write, why do we need a dedicated reader? The main argument I can see for this is safety. You know at compile time that your cache will never be written to. Is that important enough for added complexity?

If we can do want to go this direction I think I'd prefer to have a `CacheAdvance` that reads and writes (to keep the common use case simple) and then a `CacheAdvanceReader` that only reader. In this way we wouldn't publicly expose `CacheReader`. I think there's value in having an implementation detail layer (e.g. `CacheReader`) below our public classes. Gives us more flexibility down the road.

To set expectations, I'm going to put off a close review of the code until we've aligned on the high level goals and gotten https://github.com/dfed/CacheAdvance/pull/37 ready to merge.

dfed:
Closing this out for now. We can revisit this down the line if a more semantic "read-only" version feels necessary.

PR #39, opened 2020-04-13
Update version to 1.0

dfed:
Airbnb has been using CacheAdvance in production for months now. The only outstanding feature request we have can be accomplished via an additive change.

I think it's time to call this version 1.0.

### DIFF
--- a/CacheAdvance.podspec
+++ b/CacheAdvance.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '0.3.0'
+  s.version  = '1.0.0'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'A cache that enables the performant persistence of individual messages to disk'
   s.homepage = 'https://github.com/dfed/CacheAdvance'
--- a/Package.swift
+++ b/Package.swift
@@ -39,4 +39,3 @@ let package = Package(
     ],
     swiftLanguageVersions: [.v5]
 )
-let version = Version(0, 3, 0)
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ To install CacheAdvance in your iOS project with [Swift Package Manager](https:/
 
 ```swift
 dependencies: [
-    .package(url: "https://github.com/dfed/CacheAdvance", from: "0.3.0"),
+    .package(url: "https://github.com/dfed/CacheAdvance", from: "1.0.0"),
 ]
 ```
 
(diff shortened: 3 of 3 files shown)


Package.swift
```diff
@@ -39,4 +39,3 @@ let package = Package(
     ],
     swiftLanguageVersions: [.v5]
 )
-let version = Version(0, 3, 0)
```
dfed, line 42: This line isn't needed. SPM works with the current tag (there is no mention of this field in the current [Package documentation](https://developer.apple.com/documentation/swift_packages/package)). Thank you @GeorgeLyon for teaching me this!

CacheAdvance.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '0.3.0'
+  s.version  = '1.0.0'
```
dfed, line 3: This field is [required](https://guides.cocoapods.org/syntax/podspec.html#version).

dfed:
It looks like codecov has changed how it views whitespace. The delta is incorrect – I will merge through the CI block once this PR is approved.

bachand:
> It looks like codecov has changed how it views whitespace.

Can you help me understand how you determined this? I have a lot of trouble reading the output of this tool in general. Curious why it only applied to a single file?

dfed:
I clicked through on the [Sources/CacheAdvance/CacheReader.swift](https://codecov.io/gh/dfed/CacheAdvance/pull/39/diff?src=pr&el=tree#diff-U291cmNlcy9DYWNoZUFkdmFuY2UvQ2FjaGVSZWFkZXIuc3dpZnQ=) link, clicked [Coverage Changes](https://codecov.io/gh/dfed/CacheAdvance/pull/39/changes), and then opened up the CacheReader.swift diff and saw this:
image

This looks explicitly like whitespace. When I tap an "expand" button between lines, I see code I recognize, and the line that was previously displayed is blank:
image

Additionally, the diff page shows `-11` for this page:
image

And the number of lines highlighted in the first screenshot add up to 11. At which point, I'm pretty convinced that what's happening here is that codecov has a whitespace bug.

There's also the facts that:
1. 4 other files appear on the "Coverage Changes" page, each with whitespace or open brackets.
1. Only one of the 5 files modified shows up as having a coverage delta.

This leads me to think that codecov has a bug. My statement that "codecov has changed how it views whitespace" wasn't a full analysis, but given that I changed no source in this PR and codecov was still reporting a drop in code coverage, I had skipped directly to "this is a bug that I shouldn't care that much about" 😅

dfed:
I also got weird output on a similar PR in Valet: https://github.com/square/Valet/pull/223#issuecomment-612713663. Seems like codecov is having some issues with PRs that don't change code? 🤔

bachand:
Thanks for taking the time to explain.
Clicking into codecov from the Valet PR it's also weird that it seems to think that there are no changes, despite what it said in the PR comment.

Screen Shot 2020-04-18 at 6 29 17 PM

Bizarre.

PR #41, opened 2020-04-24
Add performance tests to CacheAdvance

dfed:
This PR should be reviewed using the Files changed view.

This PR introduces performance tests to CacheAdvance, as well as tests that compare CacheAdvance's performance to SQLite. The results are that CacheAdvance is more than 15x faster at appending messages than SQLite. SQLite is expectedly _significantly_ faster at reading messages, but CacheAdvance has always optimized for writing speed over reading speed.

Note that the 15x faster number is a lower bound. I've seen test runs where CacheAdvance performs more than 50x faster than SQLite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![License](https://img.shields.io/cocoapods/l/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 [![Platform](https://img.shields.io/cocoapods/p/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 
-A performant cache for logging systems. CacheAdvance persists log events 15x faster while utilizing 40% less memory than SQLite.
+A performant cache for logging systems. CacheAdvance persists log events 30x faster than SQLite.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![License](https://img.shields.io/cocoapods/l/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 [![Platform](https://img.shields.io/cocoapods/p/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 
-A cache that enables the performant persistence of individual messages to disk.
+A cache that enables the performant persistence of individual messages to disk. CacheAdvance message writes are more than 15x faster than raw SQLite writes.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![License](https://img.shields.io/cocoapods/l/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 [![Platform](https://img.shields.io/cocoapods/p/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 
-A cache that enables the performant persistence of individual messages to disk. CacheAdvance message writes are more than 15x faster than raw SQLite writes.
+A performant cache for logging systems. CacheAdvance persists log events 15x faster while utilizing 40% less memory than SQLite.
 
 ## Usage
 
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -573,7 +573,6 @@ final class CacheAdvanceTests: XCTestCase {
             version: version)
     }
 
-
     private func createCache(
         sizedToFit messages: [TestableMessage] = LorumIpsum.messages,
         overwritesOldMessages: Bool,
@@ -614,6 +613,4 @@ final class CacheAdvanceTests: XCTestCase {
     }
 
     private let testFileLocation = FileManager.default.temporaryDirectory.appendingPathComponent("CacheAdvanceTests")
-
-
 }
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -420,6 +420,62 @@ final class CacheAdvanceTests: XCTestCase {
         }
     }
 
+    func test_performance_append_fillableCache() throws {
+        measure {
+            guard let sut = try? createCache(overwritesOldMessages: false, zeroOutExistingFile: true) else {
+                XCTFail("Could not create cache")
+                return
+            }
+            for message in LorumIpsum.messages {
+                try? sut.append(message: message)
+            }
+        }
+    }
+
+    func test_performance_messages_fillableCache() throws {
+        guard let sut = try? createCache(overwritesOldMessages: false, zeroOutExistingFile: true) else {
+            XCTFail("Could not create cache")
+            return
+        }
+        for message in LorumIpsum.messages {
+            try? sut.append(message: message)
+        }
+        measure {
+            guard (try? sut.messages()) != nil else {
+                XCTFail("Could not read messages")
+                return
+            }
+        }
+    }
+
+    func test_performance_append_overwritingCache() throws {
+        measure {
+            guard let sut = try? createCache(overwritesOldMessages: true, zeroOutExistingFile: false) else {
+                XCTFail("Could not create cache")
+                return
+            }
+            for message in LorumIpsum.messages {
+                try? sut.append(message: message)
+            }
+        }
+    }
+
+    func test_performance_messages_overwritingCache() throws {
+        guard let sut = try? createCache(overwritesOldMessages: true, zeroOutExistingFile: false) else {
+            XCTFail("Could not create cache")
+            return
+        }
+        for message in LorumIpsum.messages {
+            try? sut.append(message: message)
+        }
+        measure {
+            guard (try? sut.messages()) != nil else {
+                XCTFail("Could not read messages")
+                return
+            }
+        }
+    }
+
     // MARK: Private
 
     private func requiredByteCount<T: Codable>(for messages: [T]) throws -> UInt64 {
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -420,12 +420,35 @@ final class CacheAdvanceTests: XCTestCase {
         }
     }
 
+    // MARK: Performance Tests
+
     func test_performance_append_fillableCache() throws {
+        let maximumBytes = Bytes(Double(try requiredByteCount(for: LorumIpsum.messages)))
+        // Create a cache that won't run out of room over multiple test runs
+        guard let sut = try? createCache(maximumByes: maximumBytes * 10, overwritesOldMessages: false, zeroOutExistingFile: true) else {
+            XCTFail("Could not create cache")
+            return
+        }
         measure {
-            guard let sut = try? createCache(overwritesOldMessages: false, zeroOutExistingFile: true) else {
-                XCTFail("Could not create cache")
-                return
+            // Append all messages, filling up the cahce.
+            for message in LorumIpsum.messages {
+                try? sut.append(message: message)
             }
+        }
+    }
+
+    func test_memory_append_fillableCache() throws {
+        guard #available(iOS 13.0, tvOS 13.0, macOS 10.15, *) else {
+            return
+        }
+        let maximumBytes = Bytes(Double(try requiredByteCount(for: LorumIpsum.messages)))
+        // Create a cache that won't run out of room over multiple test runs
+        guard let sut = try? createCache(maximumByes: maximumBytes * 10, overwritesOldMessages: false, zeroOutExistingFile: true) else {
+            XCTFail("Could not create cache")
+            return
+        }
+        measure(metrics: [XCTMemoryMetric()]) {
+            // Append all messages, filling up the cahce.
             for message in LorumIpsum.messages {
                 try? sut.append(message: message)
             }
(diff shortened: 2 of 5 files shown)


Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,112 @@
...
+            return
+        }
+        guard sqlite3_step(createTableStatement) == SQLITE_DONE else {
+            XCTFail("Failed to create database")
+            return
+        }
+        sqlite3_finalize(createTableStatement)
+    }
+
+    func insertMessage(_ message: String) {
```
dfed, line 75: Note that this insert isn't nearly as smart as ours. It isn't trying to limit the size of the on-disk cache, and does no overwriting of existing messages.

README.md
```diff
@@ -7,7 +7,7 @@
 [![License](https://img.shields.io/cocoapods/l/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 [![Platform](https://img.shields.io/cocoapods/p/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 
-A cache that enables the performant persistence of individual messages to disk.
+A cache that enables the performant persistence of individual messages to disk. CacheAdvance message writes are more than 15x faster than raw SQLite writes.
```
bachand, line 10: What is a non-raw SQLite write?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -420,6 +420,62 @@ final class CacheAdvanceTests: XCTestCase {
         }
     }
 
+    func test_performance_append_fillableCache() throws {
+        measure {
+            guard let sut = try? createCache(overwritesOldMessages: false, zeroOutExistingFile: true) else {
```
bachand, line 425: Should we zero out the file before the test starts?

bachand: Or should we write a new file each time so that doesn't factor into the test?

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        sqlite3_close(db)
+
+        super.tearDown()
+    }
+
+    // MARK: Behavior Tests
+
+    func test_performance_append_sqlite() throws {
+        measure {
+            createDatabase()
```
bachand, line 35: The below tests doesn't include DB creation time. This test does.
What's a bit weirder too is that when you create the DB you delete the old DB.
So if this test ran second, it would increase in time since it would need to delete the DB from the previous test. But if it's first there is no DB to delete.
Similar to my comments on the other tests I wonder if we should pull DB creation of the `measure` block.
We could even have a test for "starting up" each cache if we wanted to measure that separately.

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    func insertMessage(_ message: String) {
+        let insertStatementString = "INSERT INTO Messages(Message) VALUES ('\(message)');"
+        var insertStatement: OpaquePointer?
+
+        guard sqlite3_prepare_v2(db, insertStatementString, -1, &insertStatement, nil) == SQLITE_OK else {
+            XCTFail("Failed to prepare message")
+            return
+        }
+        sqlite3_bind_text(insertStatement, 2, NSString(string: message).utf8String, -1, nil)
```
bachand, line 83: What does "2" mean?

bachand, line 83: Do we need to explicitly bridge to `NSString`?

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        guard sqlite3_step(insertStatement) == SQLITE_DONE else {
+            XCTFail("Failed to insert message")
+            return
+        }
+        sqlite3_finalize(insertStatement)
```
bachand, line 88: It's hard to mentally validate that this code actually works 😅 
What do you think about adding a third test to verify we can actually read messages the messages out that we wrote in?

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    func readMessages() -> [String] {
+        let rawQuery = "SELECT * FROM Messages;"
+        var query: OpaquePointer?
+        guard sqlite3_prepare_v2(db, rawQuery, -1, &query, nil) == SQLITE_OK else {
```
bachand, line 94: I wish there were fewer magic numbers :(

README.md
dfed: I was trying to indicate that the SQLite writes had no logic to them beyond "insert". I'll reword since that isn't clear 😅

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: Zero-ing out the file is the same as writing a blank file to this location. The idea was that this test was for a fillable cache, which needed to start from 0 at the beginning of each test. I'll add some comments.

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
dfed: I followed some online tutorials to get this SQL code working. I'm not sure what it does 😅. We def need some number here... I'll see if I can make this more clear. But keeping this particular test code super maintainable isn't a goal of mine.

dfed: That's what the tutorials I found online did 😅. I'll see if I can find some others that don't do this. But this seemed like the easiest way to get a `UnsafePointer<Int8>` that represented a string.

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
dfed: yeah I'll add that test. I def ran that manually myself while writing this. Can't hurt to include it

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
dfed: same 😬. I'll see if I can clean this up but will time box. I don't think there's a lot of value in keeping this code terribly maintainable.

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
dfed: Good call-out. These tests are now a bit more apples-to-apples.

README.md
```diff
@@ -7,7 +7,7 @@
 [![License](https://img.shields.io/cocoapods/l/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 [![Platform](https://img.shields.io/cocoapods/p/CacheAdvance.svg)](https://cocoapods.org/pods/CacheAdvance)
 
-A cache that enables the performant persistence of individual messages to disk.
+A performant cache for logging systems. CacheAdvance persists log events 15x faster while utilizing 40% less memory than SQLite.
```
dfed, line 10: the memory brag needs more validation. I ran the tests a few times... and now the results are flipping. Investigating.

bachand: I wonder if it could vary based on lower-level decisions made by iOS? 🤔

bachand: Memory also seems like something that could vary much more than write time based on subtle differences in configuration and setup. If determining the memory performance vs SQLite proves non-obvious it could make sense to explore the comparison in a future PR. The time performance wins are a great step in and of themselves 👍

dfed: I'm actually running these tests on the Mac target to try to avoid potential iOS simulator weirdness. But I agree that the time perf wins are still excellent and worth committing.

@martinmroz is jumping on this PR to help me understand the memory delta, as I may be mis-configuring SQL for a true apples-to-apples test. Would like to do a bit more investigation before I pull the plug on that brag entirely 🙂

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
dfed: In later commits I ended up using pure Swift (rather than bridging to Objective-C types). I get why the tutorials used Objective-C types now... dealing with pointers using Swift is _rough_.

dfed:
After pulling up some of these tests in Instruments, I'm better understanding the memory differences between SQLite and CacheAdvance on a per-message-write basis.

SQLite does use create less memory on a per-write basis than CacheAdvance does. The way CacheAdvance currently writes its header is greedy – we read and write an entire header on every message write. This read and write contribute to about 1kb of transient memory usage per 500 message writes. We could likely reduce our memory footprint here, but honestly the cost of JSONEncoding is so much larger (in this case, 15x larger) than the header costs that I'm not sure optimizing here is worth the investment.

As the memory XCTests showed (before they were deleted), both CacheAdvance and SQLite had 0kb peak memory usage – effectively both of these systems have seriously low memory overhead when the caches are open and in a steady state.

I'm not entirely sure how I managed to get the 40% savings in my first memory tests. But that was clearly wrong.

bachand:
> CacheAdvance and SQLite had 0kb peak memory usage

Can you help me understand what "peak" means? I'm trying to reconcile this sentence with the previous statements about memory usage that we see due to writing the header and due to JSON encoding.

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,307 @@
...
+            _ = cache.messages()
+        }
+    }
+
+    // MARK: Private
+
+    private let testFileLocation = FileManager.default.temporaryDirectory.appendingPathComponent("SQLiteTests")
+}
+
+class SQLiteCache<T: Codable> {
```
bachand, line 163: To be honest I didn't fully process the SQLite code since we have test coverage that it does what we expect.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -22,6 +22,15 @@ import XCTest
 
 final class CacheAdvanceTests: XCTestCase {
 
+    // MARK: XCTestCase
+
+    override func setUp() {
+        super.setUp()
+
+        // Clear out any existing cache.
+        FileManager.default.createFile(atPath: testFileLocation.path, contents: nil, attributes: nil)
```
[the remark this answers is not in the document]
dfed: huh. I guess we don't need this anymore. I liked the idea of starting every test clean but I guess we've built a more explicit system 🙂

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -461,6 +470,77 @@ final class CacheAdvanceTests: XCTestCase {
...
     }
 
+    // MARK: Performance Tests
+
+    func test_performance_createCacheAndAppendSingleMessage() throws {
+        measure {
+            // Delete any existing cache.
+            FileManager.default.createFile(atPath: testFileLocation.path, contents: nil, attributes: nil)
+
+            guard let sut = try? createCache(maximumByes: 100, overwritesOldMessages: true, zeroOutExistingFile: false) else {
```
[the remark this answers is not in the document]
dfed: We need it because sometimes we're trying to create a new cache instance that reads from a file created by another cache. In those instances, we need to set `zeroOutExistingFile` to `false`.

Tests/CacheAdvanceTests/FlushableCachePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,275 @@
...
+        for message in LorumIpsum.messages {
+            try cache.appendMessage(message)
+        }
+
+        XCTAssertEqual(try cache.messages(), LorumIpsum.messages)
+    }
+
+    // MARK: Performance Tests
+
+    func test_performance_createDatabaseAndAppendSingleMessageAndFlush() {
```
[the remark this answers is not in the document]
dfed: yeah I'll rename to `cache`. This was copy/paste from the sqlite file 😬

Tests/CacheAdvanceTests/FlushableCachePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,275 @@
...
+        // We delete the database between runs because this cache type becomes less performant as the cache grows.
```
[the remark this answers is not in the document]
dfed: will follow up down on https://github.com/dfed/CacheAdvance/pull/41#discussion_r415474959

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,307 @@
...
+                location: testFileLocation,
+                maxMessageCount: Int32(LorumIpsum.messages.count),
+                shouldOverwriteMessages: false)
+
+            cache.appendMessage("test message")
+        }
+    }
+
+    func test_performance_sqlite_append_fillableCache() {
+         // Create a cache that won't run out of room over multiple test runs.
```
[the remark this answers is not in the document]
dfed: `measure` is called 10 times for wall-clock tests. For memory tests, it's called 5 times. Other tests may yet be different, but I haven't confirmed.

`measure` has called the block 10 times for as long as I've been using this function (at least 6 years). I wish this were documented rather than observed, but, it is running 10 times and then producing an average. You can see this in the CI output. Example (from a local run):
```
Test Case '-[CacheAdvanceTests.CacheAdvanceTests test_performance_messages_overwritingCache]' measured [Time, seconds] average: 0.007, relative standard deviation: 7.324%, values: [0.007642, 0.006779, 0.006662, 0.007822, 0.007862, 0.006959, 0.006448, 0.006547, 0.006742, 0.006727], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
```

Tests/CacheAdvanceTests/FlushableCachePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,275 @@
...
+            try? cache.appendMessage("test message")
+            try? cache.flushMessages()
+        }
+    }
+
+    func test_performance_flushableCache_appendAndFlush_fillableCache() {
+         // We delete the database between runs because this cache type becomes less performant as the cache grows.
+        measure {
+            // Delete the existing cache.
+            FileManager.default.createFile(atPath: testFileLocation.path, contents: nil, attributes: nil)
```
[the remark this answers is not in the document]
dfed: We can't create the cache outside of `measure` here due to the comment above the `measure` block. I think the key here is that `measure` runs multiple times. With that knowledge, I hope this test + the comment above makes much more sense.

Tests/CacheAdvanceTests/SQLitePerformanceComparisonTests.swift
```diff
@@ -0,0 +1,307 @@
...
+
+    func appendMessage(_ message: T) {
+        createDatabaseIfNecessay()
+        guard let messageData = try? encoder.encode(message),
+            let messageDataPointer = messageData.withUnsafeBytes({ $0.baseAddress }) else {
+            XCTFail("Could not encode message")
+            return
+        }
+
+        if messageCount() == maxMessageCount {
```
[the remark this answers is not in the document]
dfed: good call. I'll add this test both here and to the flushing cache.

Tests/CacheAdvanceTests/FlushableCachePerformanceComparisonTests.swift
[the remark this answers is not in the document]
dfed: Actually I can pull this logic out by telling the cache to drop the in-memory store 🤔

dfed:
> Can you help me understand what "peak" means? I'm trying to reconcile this sentence with the previous statements about memory usage that we see due to writing the header and due to JSON encoding.

Here's my guess based on observations made while writing and running these tests: "peak memory" is memory allocated during the test, and kept alive at the end of the test. Effectively, it's memory that is allocated but not released during a test run.

That said, I could very well be wrong. "Peak memory" isn't defined anywhere ([the documentation](https://developer.apple.com/documentation/xcode/improving_your_app_s_performance/reducing_your_app_s_memory_use/preventing_memory-use_regressions) isn't explicit). And I never saw a test have peak memory above 0kb – maybe these tests had peak memory in bytes but the test only registers kb? Honestly it's all pretty confusing, and may just be buggy.

I honestly don't have a lot of confidence in the memory tests. I ended up using the Instruments app on existing tests to get a better understanding of the memory footprint of CacheAdvance vs SQLite.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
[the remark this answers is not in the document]
bachand: https://github.com/dfed/CacheAdvance/pull/45

bachand:
Cool thanks for explaining the peak memory usage stuff 👍

PR #42, opened 2020-04-24
Create CADCacheAdvance: a Objective-C compatibility wrapper around a CacheAdvance<Data>

dfed:
This PR creates a new target within the CacheAdvance package: `CADCacheAdvance`. This wrapper is a very thin, Objective-C compatible wrapper around a `CacheAdvance<Data>`.

I'm open to naming suggestions, as well as API suggestions. Note that you can't both `throw` and return a `Bool` in an `@objc` method.

Note that the unit test coverage is meant to do nothing more than exercise `CADCacheAdvance`'s API in Objective-C – we already have full coverage of the underlying `CacheAdvance<T>` in the `CacheAdvanceTests` target.

cc @martinmroz


Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// - Parameter message: A message to write to disk. Must be smaller than both `maximumBytes - FileHeader.expectedEndOfHeaderInFile` and `MessageSpan.max`.
+    @objc
+    public func appendMessage(_ message: Data) throws {
+        try cache.append(message: message)
+    }
+
+    /// - Returns: `true` when there are no messages written to the file, or when the file can not be read.
+    @objc
+    public var isEmpty: Bool {
+        (try? cache.isEmpty()) ?? true
```
dfed, line 76: I don't love that failing to determine if the cache is empty leads us to say that the cache is empty. But the fact that I can't `throw` and return a `Bool` in an `@objc` method has me short on ideas.

If we can't read the file, it's as good as empty... but that's a bit flimsy. Curious to hear the thoughts of reviewers.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,88 @@
...
+    @objc
+    public var fileURL: URL {
+        cache.fileURL
+    }
+
+    /// Checks if the all the header metadata provided at initialization matches the persisted header. If not, the cache is not writable.
+    /// - Returns: `true` if the cache is writable.
+    @objc
+    public var isWritable: Bool {
+        (try? cache.isWritable()) ?? false
```
dfed, line 63: If we can't determine if a file is writable, then it isn't writable. Handling errors in Objective-C is not great, so consolidating to a simple `Bool` answer seems reasonable.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// - Warning: `maximumBytes` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will prevent appending new messages to this cache.
+    /// - Warning: `shouldOverwriteOldMessages` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will prevent appending new messages to this cache.
+    @objc
+    public init(
+        fileURL: URL,
+        maximumBytes: Bytes,
+        shouldOverwriteOldMessages: Bool)
+        throws
+    {
+        cache = try CacheAdvance<String>(
```
dfed, line 46: we're using `String` under the hood, even though we're accepting `Data` because `NSData` and `Codable` [don't work well together](https://travis-ci.com/github/dfed/CacheAdvance/jobs/323137784#L1908). Sample CI error we got when trying to store Data that had been encoded and passed into this object (linked in the prior sentence):
```
"Error Domain=NSCocoaErrorDomain Code=4866 "Top-level Data encoded as string JSON fragment." UserInfo={NSDebugDescription=Top-level Data encoded as string JSON fragment., NSCodingPath=(
1909)}"
```

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,124 @@
...
+    /// - Warning: `maximumBytes` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will prevent appending new messages to this cache.
+    /// - Warning: `shouldOverwriteOldMessages` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will prevent appending new messages to this cache.
+    @objc
+    public init(
+        fileURL: URL,
+        maximumBytes: Bytes,
+        shouldOverwriteOldMessages: Bool)
+        throws
+    {
+        cache = try CacheAdvance<Storage>(
```
dfed, line 46: We're using `Storage` under the hood, even though we're accepting `Data` and `String` because raw `NSData` and `NSString` don't work well with `Codable`. [Sample CI error we got when trying to store Data that had been encoded and passed into this object](https://travis-ci.com/github/dfed/CacheAdvance/jobs/323137784#L1908):
```
"Error Domain=NSCocoaErrorDomain Code=4866 "Top-level Data encoded as string JSON fragment." UserInfo={NSDebugDescription=Top-level Data encoded as string JSON fragment., NSCodingPath=(
1909)}"
```

Sources/CADCacheAdvance/CADCacheAdvance.swift
martinmroz: This is a reasonable compromise, I don't think this returning `true` would cause you to take any destructive action.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,124 @@
...
+        try cache.append(message: Storage(data))
+    }
+
+    /// - Returns: `true` when there are no messages written to the file, or when the file can not be read.
+    @objc
+    public var isEmpty: Bool {
+        (try? cache.isEmpty()) ?? true
+    }
+
+    /// Fetches all messages from the cache.
```
martinmroz, line 86: `/// Fetches and removes...`?

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,124 @@
...
+    @objc
+    public func messageData() throws -> [Data] {
+        try cache.messages().map { $0.d }
+    }
+
+    /// Fetches all messages from the cache.
```
martinmroz, line 92: Ditto

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,124 @@
...
+    fileprivate init(_ data: Data) {
+        d = data
+    }
+
+    // MARK: Fileprivate
+
+    fileprivate let d: Data
+
+    fileprivate var asString: String {
+        String(decoding: d, as: UTF8.self)
```
martinmroz, line 122: I think a base64 encoding would be more flexible in the event there are `null` bytes in the data. I think String may act poorly (panic?) if the data provided isn't valid UTF-8.

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: Retrieving messages doesn't remove the messages.

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: also ditto 😉

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: The issue with base64 encoding is that there's is no non-optional api to convert from String to base64 Data. So we'd have to `throw` on failure to encode.

I'm thinking it may make sense to just get rid of the `String` methods entirely, and let folk encode to their own liking. Thoughts @martinmroz?

CacheAdvance.podspec
```diff
@@ -1,13 +1,13 @@
...
-  s.version  = '1.0.1'
+  s.version  = '1.1.0'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'A cache that enables the performant persistence of individual messages to disk'
   s.homepage = 'https://github.com/dfed/CacheAdvance'
   s.authors  = 'Dan Federman'
   s.source   = { :git => 'https://github.com/dfed/CacheAdvance.git', :tag => s.version }
   s.swift_version = '5.1'
-  s.source_files = 'Sources/CacheAdvance/**/*.{swift}', 'Sources/SwiftTryCatch/**/*.{h,m}'
+  s.source_files = 'Sources/**/*.{swift}', 'Sources/**/*.{h,m}'
```
bachand, line 10: Should we make subspecs so that not everyone needs to pull down the ObjC wrapper if they don't need it?

Scripts/build.swift
```diff
@@ -117,6 +117,7 @@ for rawPlatform in rawPlatforms {
         "-configuration", "Release",
         "-derivedDataPath", platform.derivedDataPath,
         "-PBXBuildsContinueAfterErrors=0",
+        "OTHER_CFLAGS='-DGENERATED_XCODE_PROJECT'",
```
bachand, line 120: What does this do?

bachand: And why is it necessary? (Asking to learn)

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,127 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#if SWIFT_PACKAGE_MANAGER
+import CacheAdvance
+#endif
+import Foundation
+
+/// A cache that enables the performant persistence of individual messages to disk.
```
bachand, line 23: It's a bummer we will need to keep comments like this in sync but it probably is good to add them here too.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,127 @@
...
+/// This cache is intended to be written to and read from using the same serial queue.
+/// - Attention: This type is meant to be used by Objective-C code, and is not exposed to Swift. Swift code should use CacheAdvance<T>.
+@objc(CADCacheAdvance)
+@available(swift, obsoleted: 1.0)
```
bachand, line 27: 👍

Sources/CADCacheAdvance/CADCacheAdvance.swift
bachand: I think I've suggested making this change in Swift too. I think there's a valid POV that it's either writable or it isn't, and if it isn't it doesn't matter why it isn't.
But I also like not hiding errors.
I wonder if there is a way where we can make some of the Swift methods a bit more ergonomic while still giving you an option to see the full reasons (by calling into a more verbose method that throws) if you want.
But this is for another day.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Checks if the all the header metadata provided at initialization matches the persisted header. If not, the cache is not writable.
+    /// - Returns: `true` if the cache is writable.
+    @objc
+    public var isWritable: Bool {
+        (try? cache.isWritable()) ?? false
+    }
+
+    /// Appends a message to the cache.
+    /// - Parameter message: A message to write to disk. Must be smaller than both `maximumBytes - FileHeader.expectedEndOfHeaderInFile` and `MessageSpan.max`.
```
bachand, line 69: I didn't validate all of these comments. I assume you copied them well 👍

bachand: I also didn't validate that you copied all of the public methods over.

Sources/CADCacheAdvance/CADCacheAdvance.swift
bachand: I think that's part of the beauty of the overwriting version. You don't need to remove since eventually you'll just overwrite.

Sources/CADCacheAdvance/CADCacheAdvance.swift
bachand: Do we know what OS version this starts to become a problem on? It could be nice to document that so that we can more easily remove this `Storage` class when we drop support for older versions.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,127 @@
...
+        d = Data(string.utf8)
+    }
+
+    fileprivate init(_ data: Data) {
+        d = data
+    }
+
+    // MARK: Fileprivate
+
+    /// The underlying stored data. This property name is short to reduce the required on-disk storage space required per message.
```
bachand, line 121: 👌

Sources/CADCacheAdvance/CADCacheAdvance.swift
bachand: I looked into this once. If it finds invalid data I believe it uses a replacement character.
That being said, I like the idea of removing the string helpers and letting consumers decide what to do.
It's easy to write a category.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,127 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#if SWIFT_PACKAGE_MANAGER
+import CacheAdvance
```
bachand, line 19: Why does the generated Xcode project not need this?

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -0,0 +1,107 @@
...
+#import <CADCacheAdvance/CADCacheAdvance-Swift.h>
+#import <XCTest/XCTest.h>
+#else
+// These imports work when working within Package.swift.
+@import CADCacheAdvance;
+@import XCTest;
+#endif
+
+/// Tests that exercise the API of CADCacheAdvance.
+/// - Note: Since CADCacheAdvance is a thin wrapper on CacheAdvance<Data>, these tests are intended to do nothing more than exercise the API.
```
bachand, line 29: 👍

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -0,0 +1,107 @@
...
+@interface CADCacheAdvanceTests : XCTestCase
+@end
+
+@implementation CADCacheAdvanceTests
+
+- (void)test_fileURL_returnsUnderlyingFileURL;
```
bachand, line 35: I don't think we need these trailing semicolons since we're defining the method, not declaring it.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -0,0 +1,107 @@
...
+
+- (void)test_messageDataAndReturnError_returnsAppendedMessage;
+{
+    CADCacheAdvance *const cache = [self createCache];
+    NSData *const data = [@"Test" dataUsingEncoding:NSUTF8StringEncoding];
+    NSError *error = nil;
+    [cache appendData:data
+                error:&error];
+    XCTAssertNil(error);
+    XCTAssertEqualObjects([cache messageDataAndReturnError:nil], @[data]);
```
bachand, line 71: It's a confusing auto-generated method name but I guess it's idiomatic?

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,127 @@
...
+
+    /// - Returns: `true` when there are no messages written to the file, or when the file can not be read.
+    @objc
+    public var isEmpty: Bool {
+        (try? cache.isEmpty()) ?? true
+    }
+
+    /// Fetches all messages from the cache.
+    @objc
+    public func messageData() throws -> [Data] {
```
bachand, line 90: messagesData? (plural)

Sources/CADCacheAdvance/CADCacheAdvance.swift
bachand: If we go this path let's simplify the method names and documentation for the data ones that we have now 👍

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: I believe macOS 10.15, tvOS/iOS/watchOS 13 is the first OS where this starts working.

However, since we're persisting on-disk I don't think we'll ever be able to change the on-disk format without rolling a breaking change. Breaking changes that prevent reading old caches are pretty rough for consumers of this module.

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: I did copy 🙂

Scripts/build.swift
dfed: This enables us to check `#if GENERATED_XCODE_PROJECT` in code. It enables the following code:
https://github.com/dfed/CacheAdvance/blob/345ca001533be5599aa2e000ea4a9134fbb0b1da/Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m#L18-L26

We need this because the generated Xcode project requires different import styles than required by compiling when opening up the  `Package.swift`. This is a rough edge of Swift Package Manager that I ran into in this PR.

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: Simplified to just a Data API in https://github.com/dfed/CacheAdvance/pull/42/commits/bd190750a11fd01a582dd969960ccae63d385ebf

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: My thinking as well

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: The generated project actually does get this import. `SWIFT_PACKAGE_MANAGER` is defined in `Package.swift`:
https://github.com/dfed/CacheAdvance/blob/345ca001533be5599aa2e000ea4a9134fbb0b1da/Package.swift#L28

This setting applies to any Xcode project that is derived from the `Package.swift`, i.e. an Xcode project created by double-clicking the `Package.swift`, or a generated package created via `swift package generate-xcodeproj`. Basically any Xcode project created by Swift Package Manager gets this flag.

However, our Cocoapods release doesn't need this flag. Cocoapods ships a single library, rather than a separate library for `CADCacheAdvance`, `CacheAdvance`, and `SwiftTryCatch`. Therefore, we can't include `import` statements between our SPM libraries in our Cocoapods package or we'd introduce a compiler error.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
dfed: We don't need them. It's a holdover from my time at Square. Square's Objective-C style guide had every method having a `;` at the end of the line, even in implementation files. This was to aid in the copy/pasting method signatures between .m and .h files. It stuck with me 🙂

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
dfed: yeah I don't love the name, but it is what Apple is giving us. Curious what @martinmroz (who may actually use `CADCacheAdvance`) thinks of this API? I'm unlikely to roll a breaking version change in order to fixup the Objective-C API, so you better be happy with it before we merge 🙂

Sources/CADCacheAdvance/CADCacheAdvance.swift
bachand: Idea: what do you think of constructing a _new_ `Data` object and reading the data from the object passed to us from `Objective-C` into the new `Data` object? Presumably this new `Data` object (constructed in Swift and *not* bridged from `NSData`) will work fine with `Codable` on all OS versions.

I see a few advantage of this approach:
1. Probably less code.
2. We _hopefully_ won't need to have a `Storage` class around forever. I don't fully understand the problem but once we only support macOS 10.15+ and tvOS/iOS/watchOS 13+ it _seems_ like we could remove the logic to build a new `Data` object and things would just work?

CacheAdvance.podspec
dfed: I'm tempted to avoid the complication of subspecs. The `CADCacheAdvance` code is pretty lightweight (won't add much to binary size), and it can be stripped by folk using static libraries who don't use it.

We can always improve by adding subspecs down the line, so we don't lose much by bundling everything together today.

Scripts/build.swift
bachand: Good to know.
Asking to learn: why do we generate the Xcode project and build that way, vs. building using SPM?
Do we have any CI tests that verify the SPM builds directly (i.e. that don't generate the Xcode project)? I didn't seen any when I briefly looked.

dfed: I don't know how to use the command line to run-and-build on specific simulators using SPM directly, so I generate the project and use `xcodebuild` (which I'm pretty familiar with). It's possible we could simplify a good bit by relying on `swift` rather than `xcodebuild`, but I don't know and the ROI of finding out seems small?

I do not have any tests that directly test SPM. That could make a good addition.

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: I tried this. Unfortunately, we still end up with the error that we’re trying to encode a top-level object. Basically JSONEncoder on these older operating systems requires a JSON-compatible type. `Codable` alone doesn’t cut it. I had previously run into a similar error when writing unit tests, which is why `TestableMessage` exists.

dfed:
I’ve added a commit that should capture the test coverage in CADCacheAdvance. I don’t know why it isn’t being picked up. Given that I am only exercising these methods to see what the API looks like, I’m okay merging this PR even if we don’t get coverage numbers for the new target.

Scripts/build.swift
bachand: https://github.com/dfed/CacheAdvance/issues/44

Sources/CADCacheAdvance/CADCacheAdvance.swift
bachand: I wonder if CocoaPods subspecs would avoid the need for this?

bachand: A code comment explaining this would be helpful.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -0,0 +1,96 @@
...
+//  distributed under the License is distributed on an "AS IS"BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#if GENERATED_XCODE_PROJECT
+// These imports work when working within the generated CacheAdvance.xcodeproj used in CI.
+#import <CADCacheAdvance/CADCacheAdvance-Swift.h>
+#import <XCTest/XCTest.h>
```
bachand, line 21: I kind of wonder if we could use `@import XCTest` in both cases.
The `-Swift.h` makes more sense as something that needs to be different.

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: My understanding (which could be wrong) is that subspecs conditionally include code, but don't make new modules/libraries/targets. As such, I don't think subspecs would help. But def worth adding a comment. Will do.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
dfed: I believe we can. I thought there was some benefit to symmetry but maybe not. Updated.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,108 @@
...
+    /// - Returns: `true` if the cache is writable.
+    @objc
+    public var isWritable: Bool {
+        (try? cache.isWritable()) ?? false
+    }
+
+    /// Appends a message to the cache.
+    /// - Parameter message: A message to write to disk. Must be smaller than both `maximumBytes - FileHeader.expectedEndOfHeaderInFile` and `MessageSpan.max`.
+    @objc
+    public func appendMessage(_ message: Data) throws {
```
bachand, line 73: This is a bit off topic from this PR, but I'm going to add the comment anyway since this PR made me think of it.
Do we expect this data to be String? Or can it truly be any raw stream of bytes?
I guess I'm curious how non-string data gets encoded to JSON.
It seems to work but I'm curious how it's represented...

bachand: (Though I guess I'm not sure we have any tests for non-UTF-8 data? It could be interesting to add one just to be sure, if we haven't explicitly tested that use case anywhere.)

Sources/CADCacheAdvance/CADCacheAdvance.swift
bachand: I think you're probably right re: subspecs

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
bachand: We're definitely splitting hairs at this point haha

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -16,14 +16,13 @@
...
+// This import works when working within the generated CacheAdvance.xcodeproj used in CI.
 #import <CADCacheAdvance/CADCacheAdvance-Swift.h>
-#import <XCTest/XCTest.h>
 #else
-// These imports work when working within Package.swift.
+// This import works when working within Package.swift.
 @import CADCacheAdvance;
-@import XCTest;
 #endif
+@import XCTest;
```
bachand, line 25: I do like the idea of symmetry but I think this also makes a bit more clear why we are diverging in that one case. Before I wasn't immediately sure if it had to do with `#import` vs. `@import`.

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: This `message` could be anything. It's likely a string, but it could easily be any serialized type. Once we have an object as Data, we treat it as Data (we do not assume anything about it, except that it can be put in a Codable object and then serialized as JSON).

Given my performance measurement work on the other branch, I'm starting to wonder if I need to take a different approach to this PR.

The thin Objective-C wrapper on top of CacheAdvance is great, in that it minimizes the maintenance burden for me. However, it does mean that messages sent to `CADCacheAdvance` are likely being encoded twice. Once `Data` so that it can be sent to `CADCacheAdvance`, and then again as `Storage` so that `CacheAdvance` can persist them.

I may need to offer a way to customize serialization/deserialization code within `CacheAdvance` (at least internally probably not worth making that API public). Going to sit on this thought for a bit.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
dfed: Makes a lot of sense. Reducing the need for a close reading is a good strategy for long-term maintenance (and for enabling others to contribute!).

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: Just pushed up performance tests (which required duplicating the lorum ipsum array 😞). `test_performance_append_fillableCache` for `CADCacheAdvance` is 0.06s slower (0.017s) than the same test for `CacheAdvance` (0.011s).

That feels worth fixing.

dfed: Sigh... to customize this behavior I will need to make the new API public, because `CADCacheAdvance` lives in a separate library 😫

dfed: Okay okay we've got 2910f9ddbce71e2f4ecba10123e9e42d4532dbf4 now. `CADCacheAdvance`'s `test_performance_append_fillableCache` is now taking 0.011s.

dfed: One nicety about the new public API is that if someone has a more performant way to serialize their messages (i.e. using something other than JSON encoding/decoding), they can now use it. Feels reasonable to let our consumers eek out as much performance from our system as they are able.

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -0,0 +1,127 @@
...
```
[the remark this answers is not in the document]
dfed: Ahhhh this is a good question! I think you're right we may have had issues here. The good news is that my recent commits obviate the need for JSON encoding in `CADCacheAdvance` entirely. Let me update the README 🙂

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -0,0 +1,616 @@
...
+    CADCacheAdvance *const cache = [self createCacheThatOverwitesOldMessages:YES];
+    XCTAssertTrue(cache.isEmpty);
+}
+
+- (void)test_isEmpty_returnsFalseForANonEmptyCache;
+{
+    CADCacheAdvance *const cache = [self createCacheThatOverwitesOldMessages:YES];
+    NSError *error = nil;
+    [cache appendMessage:[@"Test" dataUsingEncoding:NSUTF8StringEncoding]
+                error:&error];
```
bachand, line 59: Indentation looks off here.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -0,0 +1,616 @@
...
+{
+    CADCacheAdvance *const cache = [self createCacheThatOverwitesOldMessages:YES];
+    XCTAssertTrue(cache.isEmpty);
+}
+
+- (void)test_isEmpty_returnsFalseForANonEmptyCache;
+{
+    CADCacheAdvance *const cache = [self createCacheThatOverwitesOldMessages:YES];
+    NSError *error = nil;
+    [cache appendMessage:[@"Test" dataUsingEncoding:NSUTF8StringEncoding]
```
bachand, line 58: So just to confirm what I think I know the answer to... `Codable` is not available to Objective-C which is the root reason why we need to rework things a bit?

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -0,0 +1,616 @@
...
+// MARK: Private
+
+- (CADCacheAdvance *)createCacheThatOverwitesOldMessages:(BOOL)overwritesOldMessages;
+{
+    [NSFileManager.defaultManager createFileAtPath:[self testFileLocation].path
+                                          contents:nil
+                                        attributes:nil];
+    NSError *error = nil;
+    CADCacheAdvance *const cache = [[CADCacheAdvance alloc]
+                                    initWithFileURL:self.testFileLocation
```
bachand, line 117: Could be nice to double check this indentation. I was expecting one more space but I could very well be misremembering.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
```diff
@@ -0,0 +1,616 @@
...
+    XCTAssertNil(error, "Failed to create cache due to %@", error);
+    return cache;
+}
+
+- (NSURL *)testFileLocation;
+{
+    return [NSFileManager.defaultManager.temporaryDirectory URLByAppendingPathComponent:@"CADCacheAdvanceTests"];
+}
+
+- (NSArray<NSString *> *)lorumIpsumMessages;
```
bachand, line 130: What do you think about putting these in a file that we can read from disk in both sets of tests?
It would be nice if we had a stronger guarantee that they were in sync so that we are sure we're comparing apples to apples when we're comparing performance test results.
Another option would be to put them in their own target that we share between tests.

README.md
```diff
@@ -19,6 +19,14 @@ let myCache = try CacheAdvance<MyMessageType>(
     maximumBytes: 5000,
     shouldOverwriteOldMessages: false)
 ```
+
+```objc
+CADCacheAdvance *const cache = [[CADCacheAdvance alloc]
+                                    initWithFileURL:[NSFileManager.defaultManager.temporaryDirectory URLByAppendingPathComponent:@"MyCache"]
```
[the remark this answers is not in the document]
dfed: I think this is a GitHub line wrapping issue. This is all on one line.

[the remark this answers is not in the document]
dfed: oh oh oh you're talking about the indentation of `initWithFileURL`. Yeah I'll fix.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -29,20 +29,29 @@ public final class CacheAdvance<T: Codable> {
...
     ///   - shouldOverwriteOldMessages: When `true`, once the on-disk store exceeds maximumBytes, new entries will replace the oldest entry.
+    ///   - decoder: The decoder that will be used to decode already-persisted messages.
+    ///   - encoder: The encoder that will be used to encode messages prior to persistence.
     ///
     /// - Warning: `maximumBytes` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will prevent appending new messages to this cache.
     /// - Warning: `shouldOverwriteOldMessages` must be consistent for the life of a cache. Changing this value after logs have been persisted to a cache will prevent appending new messages to this cache.
+    /// - Warning: `decoder` must have a consistent implementation for the life of a cache. Changing this value after logs have been persisted to a cache may prevent reading messages from this cache.
+    /// - Warning: `encoder` must have a consistent implementation for the life of a cache. Changing this value after logs have been persisted to a cache may prevent reading messages from this cache.
```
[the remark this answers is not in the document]
dfed: Nope! If the encoder changes, then we might have trouble reading with the expected `decoder`. Writing will always succeed.

Sources/CacheAdvance/MessageDecoder.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+
+import Foundation
+
+/// An object capable of decoding a message of type `T` from `Data`.
+public protocol MessageDecoder {
+    func decode<T>(_ type: T.Type, from data: Data) throws -> T where T : Decodable
+}
+
+extension JSONDecoder: MessageDecoder {}
```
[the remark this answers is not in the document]
dfed: I think we can add this in a feature bump after this PR lands. Per our initial investigation of `PropertyList` vs `JSON` encoding/decoding in https://github.com/dfed/CacheAdvance/pull/1#discussion_r353587842, I'm not yet sure we want to invest time in making utilizing property list encoding easy.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
dfed: It's a couple things:
1. We do not have `Codable` available to Objective-C
1. We can not expose Swift generics to Objective-C

The second point means that even if we had `Codable` in Objective-C, we'd have no way of specifying a `<T: Codable>` generic in a way that Objective-C could use.

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
dfed: This is the correct one. But I did get it wrong in two other spots you highlighted 🙂

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
dfed: I can give that a shot

dfed: Gave it a shot in https://github.com/dfed/CacheAdvance/pull/42/commits/dc3137e7472a4c654302ef6490648329dc5f0d63. I may need to fix up some CI issues, but should be doable. Solid idea 🙂

dfed:
@martinmroz mind taking another look at the Objective-C API before this merges? Since you may be the primary consumer of the Objective-C API, I want to make sure it's in a state where you'll like using it 😉

Tests/CADCacheAdvanceTests/CADCacheAdvanceTests.m
bachand: That worked out great, and much DRY-er!

Package.swift
```diff
@@ -29,8 +29,7 @@ let package = Package(
         ),
         .testTarget(
             name: "CacheAdvanceTests",
-            dependencies: ["CacheAdvance", "LorumIpsum"],
-            swiftSettings: [.define("SWIFT_PACKAGE_MANAGER")]
```
bachand: Interesting. Is the reason that we don't need this because we don't distribute the tests via CocoaPods?

dfed: Correct!

Sources/CADCacheAdvance/CADCacheAdvance.swift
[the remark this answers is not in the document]
martinmroz: Arbitrary data should probably just be base64 encoded so we're not reliant on a separate spec to determine what can and cannot be encoded in a CacheAdvance object.

PR #45, opened 2020-04-28
Clear cache before each CacheAdvance test and remove `zeroOutExistingFile`

bachand:
_This PR is best reviewed commit by commit._

This is a follow up to this comment thread: https://github.com/dfed/CacheAdvance/pull/41#discussion_r415469848

I also like clearing the cache out before each test.

What I was trying to get out (and evidently poorly communicated) was that we could probably get rid of the `zeroOutExistingFile` argument in our cache creation methods.

This PR does that.

### DIFF
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -22,6 +22,11 @@ import XCTest
 
 final class CacheAdvanceTests: XCTestCase {
 
+    override func setUp() {
+        super.setUp()
+        clearCacheFile()
+    }
+
     // MARK: Behavior Tests
 
     func test_isEmpty_returnsTrueWhenCacheIsEmpty() throws {
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -48,7 +48,7 @@ final class CacheAdvanceTests: XCTestCase {
         let cache = try createCache(overwritesOldMessages: false)
         try cache.append(message: message)
 
-        let sameCache = try createCache(overwritesOldMessages: false, zeroOutExistingFile: false)
+        let sameCache = try createCache(overwritesOldMessages: false)
 
         XCTAssertFalse(try sameCache.isEmpty())
     }
@@ -76,9 +76,7 @@ final class CacheAdvanceTests: XCTestCase {
             version: 0)
         try originalHeader.synchronizeHeaderData()
 
-        let sut = try createCache(
-            overwritesOldMessages: false,
-            zeroOutExistingFile: false)
+        let sut = try createCache(overwritesOldMessages: false)
         XCTAssertThrowsError(try sut.isEmpty()) {
             XCTAssertEqual($0 as? CacheAdvanceError, CacheAdvanceError.incompatibleHeader)
         }
@@ -102,7 +100,7 @@ final class CacheAdvanceTests: XCTestCase {
         let originalCache = try createCache(overwritesOldMessages: false)
         XCTAssertTrue(try originalCache.isWritable())
 
-        let sut = try createCache(overwritesOldMessages: false, zeroOutExistingFile: false)
+        let sut = try createCache(overwritesOldMessages: false)
         XCTAssertTrue(try sut.isWritable())
     }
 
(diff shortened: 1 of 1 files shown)


Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -48,7 +48,7 @@ final class CacheAdvanceTests: XCTestCase {
         let cache = try createCache(overwritesOldMessages: false)
         try cache.append(message: message)
 
-        let sameCache = try createCache(overwritesOldMessages: false, zeroOutExistingFile: false)
```
bachand, line 46: We don't need this `zeroOutExistingFile` anymore since we only clear the cache explicitly with `clearCacheFile()`. Creating the cache with either of the `createCache()` variants won't clear it.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -317,6 +317,9 @@ final class CacheAdvanceTests: XCTestCase {
             let secondCacheMessages = try secondCache.messages()
```
bachand, line 317: I'm pretty sure this test had a bug before since we were overwriting the cache file when creating a second one.
I printed out the result of `secondCacheMessages` and it was only one message.
Ideally we can rework this test so it will fail if we make the same mistake again, but I didn't have time to do so.

dfed: good find!

I think the easiest solve here may be to assert that the second cache isn’t empty after creating it.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -588,7 +600,6 @@ final class CacheAdvanceTests: XCTestCase {
         throws
         -> CacheAdvance<TestableMessage>
     {
-        if zeroOutExistingFile { clearCacheFile() }
```
dfed: We should remove the argument here as well, right?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand: Good idea https://github.com/dfed/CacheAdvance/pull/45/commits/8f18f894a84c39a90f81e6a36d95e0e72d91af98

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand: Ah yes! https://github.com/dfed/CacheAdvance/pull/45/commits/8f18f894a84c39a90f81e6a36d95e0e72d91af98

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -554,9 +576,7 @@ final class CacheAdvanceTests: XCTestCase {
         throws
         -> CacheHeaderHandle
     {
-        if zeroOutExistingFile {
-            FileManager.default.createFile(atPath: testFileLocation.path, contents: nil, attributes: nil)
-        }
+        if zeroOutExistingFile { clearCacheFile() }
```
bachand, line 579: This can probably go in a future PR but I don't want to edit too much after approval 👍

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -308,10 +313,17 @@ final class CacheAdvanceTests: XCTestCase {
             let cachedMessages = try cache.messages()
 
             let secondCache = try createCache(overwritesOldMessages: true, maximumByteDivisor: maximumByteDivisor)
+
+            // Check that we can retrieve the messages from the first cache.
+            XCTAssertGreaterThan(try secondCache.messages().count, 0)
```
dfed, line 318: `XCTAssertTrue(try secondCache.isEmpty())` may be more semantic, but this absolutely works. Thanks for adding!

dfed:
All changes LGTM. Merge when green 😄

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: Feel free to delete. If it works, I’m game.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand: https://github.com/dfed/CacheAdvance/pull/45/commits/5a5f0077563719459e942e10eb09ca5f4a035ad7

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand: Gonna merge so I can move past this particular PR but hopefully I'll be back and can clean it up then.

PR #50, opened 2021-01-11
Expose encountered header version in incompatibleHeader error

dfed:
This PR enables propagating the header version of the file being read as an associated value of the `CacheAdvanceError.incompatibleHeader` error. In order to do this, I've changed the signature of `func canOpenFile(...) -> Bool` to `func checkFile(...) throws` to push error throwing down the stack to where comparisons are being done and header version data is available.

Since we're adding new information to the incompatible header error, I'm considering this to be a feature version bump. This change is technically breaking to code that creates their own `incompatibleHeader` error, but since doing that outside of `CacheAdvance` doesn't make much sense I don't think this PR warrants a major version bump.

### DIFF
--- a/CacheAdvance.podspec
+++ b/CacheAdvance.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '1.1.0'
+  s.version  = '1.2.0'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'A performant cache for logging systems. CacheAdvance persists log events 30x faster than SQLite.'
   s.homepage = 'https://github.com/dfed/CacheAdvance'
--- a/Sources/CacheAdvance/CacheAdvance.swift
+++ b/Sources/CacheAdvance/CacheAdvance.swift
@@ -73,7 +73,7 @@ public final class CacheAdvance<T: Codable> {
     /// - Parameter message: A message to write to disk. Must be smaller than both `maximumBytes - FileHeader.expectedEndOfHeaderInFile` and `MessageSpan.max`.
     public func append(message: T) throws {
         try setUpFileHandlesIfNecessary()
-        try header.openFile()
+        try header.checkFile()
         guard try header.canWriteToFile() else {
             throw CacheAdvanceError.fileNotWritable
         }
@@ -140,7 +140,7 @@ public final class CacheAdvance<T: Codable> {
     /// - Returns: `true` when there are no messages written to the file.
     public func isEmpty() throws -> Bool {
         try setUpFileHandlesIfNecessary()
-        try header.openFile()
+        try header.checkFile()
 
         return header.offsetInFileAtEndOfNewestMessage == FileHeader.expectedEndOfHeaderInFile
     }
@@ -148,7 +148,7 @@ public final class CacheAdvance<T: Codable> {
     /// Fetches all messages from the cache.
     public func messages() throws -> [T] {
         try setUpFileHandlesIfNecessary()
-        try header.openFile()
+        try header.checkFile()
 
         var messages = [T]()
         while let encodedMessage = try reader.nextEncodedMessage() {
--- a/Sources/CacheAdvance/CacheAdvance.swift
+++ b/Sources/CacheAdvance/CacheAdvance.swift
@@ -73,9 +73,7 @@ public final class CacheAdvance<T: Codable> {
     /// - Parameter message: A message to write to disk. Must be smaller than both `maximumBytes - FileHeader.expectedEndOfHeaderInFile` and `MessageSpan.max`.
     public func append(message: T) throws {
         try setUpFileHandlesIfNecessary()
-        guard try header.canOpenFile() else {
-            throw CacheAdvanceError.incompatibleHeader
-        }
+        try header.openFile()
         guard try header.canWriteToFile() else {
             throw CacheAdvanceError.fileNotWritable
         }
@@ -142,9 +140,7 @@ public final class CacheAdvance<T: Codable> {
     /// - Returns: `true` when there are no messages written to the file.
     public func isEmpty() throws -> Bool {
         try setUpFileHandlesIfNecessary()
-        guard try header.canOpenFile() else {
-            throw CacheAdvanceError.incompatibleHeader
-        }
+        try header.openFile()
 
         return header.offsetInFileAtEndOfNewestMessage == FileHeader.expectedEndOfHeaderInFile
     }
@@ -152,9 +148,7 @@ public final class CacheAdvance<T: Codable> {
     /// Fetches all messages from the cache.
     public func messages() throws -> [T] {
         try setUpFileHandlesIfNecessary()
-        guard try header.canOpenFile() else {
-            throw CacheAdvanceError.incompatibleHeader
-        }
+        try header.openFile()
 
         var messages = [T]()
         while let encodedMessage = try reader.nextEncodedMessage() {
--- a/Sources/CacheAdvance/CacheAdvanceError.swift
+++ b/Sources/CacheAdvance/CacheAdvanceError.swift
@@ -21,6 +21,7 @@ public enum CacheAdvanceError: Error, Equatable {
     /// Thrown when a message being appended to a cache that does not overwrite old messages is too large to store in the remaining space.
     case messageLargerThanRemainingCacheSize
     /// Thrown when a cache's persisted header is incompatible with the current implementation.
+    /// - Parameter persistedVersion: The header version of the file being read.
     case incompatibleHeader(persistedVersion: UInt8)
     /// Thrown when the cache file's persisted static header data is inconsistent with the metadata with which the cache was initialized.
     case fileNotWritable
--- a/Sources/CacheAdvance/CacheAdvanceError.swift
+++ b/Sources/CacheAdvance/CacheAdvanceError.swift
@@ -21,7 +21,7 @@ public enum CacheAdvanceError: Error, Equatable {
     /// Thrown when a message being appended to a cache that does not overwrite old messages is too large to store in the remaining space.
     case messageLargerThanRemainingCacheSize
     /// Thrown when a cache's persisted header is incompatible with the current implementation.
-    case incompatibleHeader
+    case incompatibleHeader(persistedVersion: UInt8)
     /// Thrown when the cache file's persisted static header data is inconsistent with the metadata with which the cache was initialized.
     case fileNotWritable
     /// Thrown when the cache file is of an unexpected format.
--- a/Sources/CacheAdvance/CacheHeaderHandle.swift
+++ b/Sources/CacheAdvance/CacheHeaderHandle.swift
@@ -77,8 +77,8 @@ final class CacheHeaderHandle {
     /// Throws if the expected header version does not match the persisted header version.
     ///
     /// - Throws: `CacheAdvanceError.incompatibleHeader` if this object's header version does not match that of `fileHeader`. May also throw a file reading error if the file can not be read.
-    func openFile() throws {
-        try openFile(with: try memoizedMetadata())
+    func checkFile() throws {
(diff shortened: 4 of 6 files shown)


Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -74,11 +74,11 @@ final class CacheHeaderHandle {
...
     }
 
-    /// Checks if the expected header version matches the persisted header version.
+    /// Throws if the expected header version does not match the persisted header version.
     ///
-    /// - Returns: `true` if this object's header version matches that of `fileHeader`; otherwise `false`.
-    func canOpenFile() throws -> Bool {
-        canOpenFile(with: try memoizedMetadata())
+    /// - Throws: `CacheAdvanceError.incompatibleHeader` if this object's header version does not match that of `fileHeader`. May also throw a file reading error if the file can not be read.
+    func openFile() throws {
```
bachand, line 80: The name of this method makes me expect it will give me a file handle back. What about `checkFile()` in the spirit of this new similar Apple API ([source](https://github.com/DougGregor/swift-evolution/blob/fa58e0c4036ea3efbf4b264592aa99fb0f2f88f8/proposals/nnnn-structured-concurrency.md))?
Screen Shot 2021-01-11 at 5 26 43 PM

Sources/CacheAdvance/CacheAdvanceError.swift
```diff
@@ -21,7 +21,7 @@ public enum CacheAdvanceError: Error, Equatable {
     /// Thrown when a message being appended to a cache that does not overwrite old messages is too large to store in the remaining space.
     case messageLargerThanRemainingCacheSize
     /// Thrown when a cache's persisted header is incompatible with the current implementation.
-    case incompatibleHeader
+    case incompatibleHeader(persistedVersion: UInt8)
```
bachand, line 24: Let's document this associated value for completeness.

Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -142,23 +144,27 @@ final class CacheHeaderHandle {
         return try handle.readDataUp(toLength: Int(FileHeader.expectedEndOfHeaderInFile))
     }
 
-    /// Checks if the expected header version matches the persisted header version.
+    /// Throws if the expected header version does not match the persisted header version.
     ///
     /// - Parameter persistedMetadata: The persisted header metadata.
-    /// - Returns: `true` if this object's header version matches that of `fileHeader`; otherwise `false`.
-    private func canOpenFile(with persistedMetadata: Metadata) -> Bool {
+    /// - Throws: `CacheAdvanceError.incompatibleHeader` if this object's header version does not match that of `fileHeader`. May also throw a file reading error if the file can not be read.
```
bachand, line 150:
```suggestion
    /// - Throws: `CacheAdvanceError.incompatibleHeader` if this object's header version does not match that of `fileHeader`.
```

Right?

Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -74,11 +74,11 @@ final class CacheHeaderHandle {
         try handle.write(data: expectedHeader.data(for: .offsetInFileAtEndOfNewestMessage))
     }
 
-    /// Checks if the expected header version matches the persisted header version.
+    /// Throws if the expected header version does not match the persisted header version.
```
bachand, line 77:
```suggestion
    /// Checks that the file is in a valid state ....<insert more?>....
```
Assuming that we go with the a name like `checkFile()` I wonder if we can make this more affirmative now and then keep the failure details in the `- Throws:` section?

Tests/CacheAdvanceTests/CacheHeaderHandleTests.swift
```diff
@@ -125,16 +125,18 @@ final class CacheHeaderHandleTests: XCTestCase {
             overwritesOldMessages: true,
             version: 2)
 
-        XCTAssertFalse(try sut.canOpenFile())
+        XCTAssertThrowsError(try sut.openFile()) {
+            XCTAssertEqual($0 as? CacheAdvanceError, CacheAdvanceError.incompatibleHeader(persistedVersion: 1))
```
bachand, line 129: Nice!

bachand (commented):
@dfed I left some comments but as I finish up my review I'm having trouble seeing the key change in this PR. Can you help me follow? My read of the code indicates that the only change in the public API is that we propagate the currently persisted file header metadata version in the `CacheAdvanceError.incompatibleHeader` error (a good thing!). But from what I can tell we always propagated file read errors. What am I missing?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -73,9 +73,7 @@ public final class CacheAdvance<T: Codable> {
     /// - Parameter message: A message to write to disk. Must be smaller than both `maximumBytes - FileHeader.expectedEndOfHeaderInFile` and `MessageSpan.max`.
     public func append(message: T) throws {
         try setUpFileHandlesIfNecessary()
-        guard try header.canOpenFile() else {
-            throw CacheAdvanceError.incompatibleHeader
-        }
+        try header.openFile()
```
dfed, line 76: @bachand the changes in this file are the ones that cause the header-reading errors to be propagated. Previously we'd swallow whatever errors were encountered when calling `header.canOpenFile`. Now we propagate those errors directly.

bachand: @dfed I feel silly posting this I feel like I must be missing something. I guess I'm not seeing what error we are swallowing. If `header.canOpenFile()` throws we propagate that error upward since the `guard` only applies if `header.canOpenFile()` returns normally.

Screen Shot 2021-01-11 at 8 42 09 PM

So I am pretty sure that if `memoizedMetadata()` (called within `canOpenFile()`) were to throw we would propagate that error outwards to the client.

What am I missing? 😖

dfed: Good writeup @bachand! So... I think I was blinded by the fact because I'm seeing reports of `CacheAdvanceError.incompatibleHeader` being thrown and so I assumed we must be squashing a better error. If we're properly propagating errors encountered... then I have absolutely no clue how we're hitting this condition.

I'm open to reverting this part of the change, since you're right it doesn't make a functional difference. I do like the idea of `CacheHeaderHandle` throwing the error since it's the code doing the comparison, but you're right that doing so isn't a requirement.

Now I'm really curious what the version we're encountering is. If I had to guess: the versions will be somewhat random, and we're encountering files that shouldn't be where we're looking.

CacheAdvance.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '1.1.0'
+  s.version  = '1.2.0'
```
bachand, line 3: We're technically changing public API in the error enum case. But maybe that's small enough that it's OK?

Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -142,23 +144,27 @@ final class CacheHeaderHandle {
         return try handle.readDataUp(toLength: Int(FileHeader.expectedEndOfHeaderInFile))
     }
 
-    /// Checks if the expected header version matches the persisted header version.
+    /// Checks that the file's persisted metadata has the expected values.
     ///
     /// - Parameter persistedMetadata: The persisted header metadata.
-    /// - Returns: `true` if this object's header version matches that of `fileHeader`; otherwise `false`.
-    private func canOpenFile(with persistedMetadata: Metadata) -> Bool {
+    /// - Throws: `CacheAdvanceError.incompatibleHeader` if this object's header version does not match that of `fileHeader`.
```
bachand, line 150: @dfed I committed the change I proposed here since it seemed like you agreed.

Sources/CacheAdvance/CacheHeaderHandle.swift
```diff
@@ -142,23 +144,27 @@ final class CacheHeaderHandle {
...
+    private func checkFile(with persistedMetadata: Metadata) throws {
         // Our current file header version is 1.
```
bachand, line 152: Is this still true? If we only have a single version then how could we be hitting this problem?

bachand (approved):
@dfed I'm glad I understand now. I'm happy to move ahead since I think this is cleaner. Let's update the PR description before merging to clarify what we're actually changing.

Sources/CacheAdvance/CacheAdvance.swift
bachand: Can you help me understand how the versions could be random? I'm not following that part.

Sources/CacheAdvance/CacheHeaderHandle.swift
dfed: Yes! Thank you for that 🙂

CacheAdvance.podspec
dfed: It's not a breaking change unless you're creating your own error... which... I don't think that's worth supporting with semver? A `catch CacheAdvanceError.incompatibleHeader` clause still works. Willing to roll a major version though since you're right that this is indeed a breaking change in some cases.

Sources/CacheAdvance/CacheHeaderHandle.swift
dfed: Good q! Will answer in this thread: https://github.com/dfed/CacheAdvance/pull/50#discussion_r555927694

Sources/CacheAdvance/CacheAdvance.swift
dfed: The following code will throw a `CacheAdvanceError.incompatibleHeader(persistedVersion: 116)`:
```
FileManager.default.createFile(
    atPath: testFileLocation.path,
    contents: "this is a somewhat long string that is not real CacheAdvance data.".data(using: .utf8)!,
    attributes: nil)
try CacheHeaderHandle(
    forReadingFrom: testFileLocation,
    maximumBytes: 5000,
    overwritesOldMessages: false).checkFile()
```

To determine the header's version, we read the first `UInt8.storageLength` of whatever file a `CacheAdvance` instance is pointed at and turn that value into an integer. If we're reading a file that wasn't written by `CacheAdvance`, then the value of the version in the file "header" is effectively random/meaningless.

dfed:
PR description updated! Waiting for explicit alignment on the versioning decision (feature bump vs breaking change) before merging.

CacheAdvance.podspec
bachand: I am OK keeping this a minor version since we wouldn't expect someone to be instantiating one of these errors.

Sources/CacheAdvance/CacheAdvance.swift
bachand: Ah got it. We may want to consider adding a unique (or as-unique-as-possible) [signature](https://en.wikipedia.org/wiki/List_of_file_signatures) value to the start of the file. That could mitigate this problem significantly. If the file doesn't have the signature value we can bail out immediately and not try to parse it.

bachand:
🚢

Sources/CacheAdvance/CacheAdvance.swift
dfed: Yup! Thankfully we reserved 38bytes in our header, so we have room to encode a signature without making a breaking change. Though if we take this approach we'd need to enable an upgrade path for files written using the version without this info. Food for future thought.

In the meantime, the version matching is effectively a signature since we know all the possible valid values. It's a short and not-very-complex signature, but it's what we've got thus far 😅

dfed: Interesting finding: the file size of the file with an incompatible header is _always_ 64bytes. Meaning that it's a file to which we never wrote any messages.

I still don't understand why the header file has an incompatible header (I screwed up and forgot to log the header version). But the good news is we're never losing messages.

PR #52, opened 2022-03-07
Prevent infinite recursion loop while reading messages

dfed:
Airbnb has seen an infinite recursion crash when reading `messages()` in the wild.

The crash's stack trace involves repeatedly calling through to `nextEncodedMessage()`. This PR makes our `nextEncodedMessage()` method defensive against an infinite recursion loop, and also fixes the underlying issue that enabled a cache file to get into this bad state. **Best reviewed commit by commit**

I will rebase this PR over #53 once that lands in order to get CI running on this PR, but I have run all tests locally and they are passing, so this should be good to review.


Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -36,20 +36,39 @@ public final class CacheAdvance<T: Codable> {
...
+            header: try CacheHeaderHandle(
+                forReadingFrom: fileURL,
+                maximumBytes: maximumBytes,
+                overwritesOldMessages: shouldOverwriteOldMessages),
+            decoder: decoder,
+            encoder: encoder)
+    }
 
+    /// An internal initializer with no logic. Can be used to create pathological test cases.
+    required init(
```
dfed, line 60: we need a way to inject these private properties so that we can mangle them prior to executing our test.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -602,7 +602,7 @@ final class CacheAdvanceTests: XCTestCase {
         throws
         -> CacheAdvance<TestableMessage>
     {
-        return try CacheAdvance<TestableMessage>(
```
dfed, line 605: saw this while I was here. This change is very much unrelated to the goal of this PR

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,6 +96,34 @@ final class CacheAdvanceTests: XCTestCase {
...
+        encoder: JSONEncoder())
+
+      // Make sure the header data is persisted before we read it as part of the `messages()` call below.
+      try header.synchronizeHeaderData()
+      // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
+      // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
+      try header.updateOffsetInFileAtEndOfNewestMessage(
+        to: FileHeader.expectedEndOfHeaderInFile + UInt64(MessageSpan.storageLength) + 1)
+
+      XCTAssertThrowsError(try cache.messages()) {
```
dfed, line 122: Before the next commit, this test will infinite loop and crash.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,6 +96,34 @@ final class CacheAdvanceTests: XCTestCase {
...
+          XCTAssertEqual($0 as? CacheAdvanceError, CacheAdvanceError.fileCorrupted)
```
dfed, line 123: We should be able to detect this case and throw an appropriate error. We should also be able to prevent this case from happening in production. We solve both of these "shoulds" in the next few commits.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,6 +96,34 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
+  func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
+      let header = try CacheHeaderHandle(
+        forReadingFrom: testFileLocation,
+        maximumBytes: 10_000,
```
dfed, line 102: this number is effectively random. Doesn't matter what it is.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -112,12 +112,16 @@ public final class CacheAdvance<T: Codable> {
...
+            let truncateAtOffset: UInt64?
+            if cacheHasSpaceForNewMessageBeforeEndOfFile {
+                // We have room for this message. No need to truncate.
+                truncateAtOffset = nil
+            } else {
                 // This message can't be written without exceeding our maximum file length.
                 // We'll need to start writing the file from the beginning of the file.
 
-                // Trim the file to the current position to remove soon-to-be-abandoned data from the file.
-                try writer.truncate(at: writer.offsetInFile)
```
dfed, line 101: I believe the way we managed to get into the state the previous commit is testing is that we were truncating the file before we updated the header. We should always update our header before making a write, not after.

I inspected other places where we write data and we were already careful in the other spots to update our header first.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -60,11 +60,17 @@ final class CacheReader {
             return message
 
         case .emptyRead:
+            guard !previousReadWasEmpty else {
+                // If the previous read was also empty, then the file has been corrupted.
+                // Two empty reads in a row means that offsetInFileAtEndOfNewestMessage is incorrect.
+                // This inconsistency likely occurred becuase a crash occured during a message write.
```
shepting, line 66:
```suggestion
                // This inconsistency likely occurred because a crash occurred during a message write.
```

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
fdiaz: let's add this to a local constant called `randomHighValue`, that way we document this via the property name.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -60,11 +60,17 @@ final class CacheReader {
             return message
 
         case .emptyRead:
+            guard !previousReadWasEmpty else {
+                // If the previous read was also empty, then the file has been corrupted.
+                // Two empty reads in a row means that offsetInFileAtEndOfNewestMessage is incorrect.
+                // This inconsistency likely occurred because a crash occurred during a message write.
```
fdiaz, line 66:
```suggestion
                // This inconsistency likely occurred because a crash happened during a message write.
```

Let's not be redundant with "occurred" 😁

fdiaz (approved):

Sources/CacheAdvance/CacheReader.swift
dfed:
```suggestion
                // This inconsistency likely is likely due to a crash occurring during a message write.
```

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: 2470113

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,6 +96,35 @@ final class CacheAdvanceTests: XCTestCase {
...
+            overwritesOldMessages: header.overwritesOldMessages),
+        decoder: JSONDecoder(),
+        encoder: JSONEncoder())
+
+      // Make sure the header data is persisted before we read it as part of the `messages()` call below.
+      try header.synchronizeHeaderData()
+      // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
+      // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
+      try header.updateOffsetInFileAtEndOfNewestMessage(
+        to: FileHeader.expectedEndOfHeaderInFile + UInt64(MessageSpan.storageLength) + 1)
```
bachand, line 121: If this file is empty would `FileHeader.expectedEndOfHeaderInFile + 1` cause the issue? I am trying to understand why we need to add `UInt64(MessageSpan.storageLength)`.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -93,12 +112,16 @@ public final class CacheAdvance<T: Codable> {
...
+                // Trim the file to the current writer position to remove soon-to-be-abandoned data from the file.
+                truncateAtOffset = writer.offsetInFile
 
                 // Set the offset back to the beginning of the file.
```
bachand, line 126: The next comment says:

> // We know the oldest message is at the beginning of the file, since we just tossed out the rest of the file.

I don't think that's accurate anymore since we haven't truncated yet.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -121,6 +144,12 @@ public final class CacheAdvance<T: Codable> {
             // If the application crashes between writing the header and writing the message data, we'll have lost the messages between the previous offsetInFileOfOldestMessage and the new offsetInFileOfOldestMessage.
             try header.updateOffsetInFileOfOldestMessage(to: offsetInFileOfOldestMessage)
```
bachand, line 145: I'm trying to load back into my mind how this code works. If the cache file allows overwriting we start to write from the beginning of the file again. It seems from the code that when we start overwriting, we seek to the next oldest message. What happens if the new message we are trying to write is so long that we need to overwrite _multiple_ older messages?

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -121,6 +144,12 @@ public final class CacheAdvance<T: Codable> {
...
 
+            // Truncate the file if it needs truncation before we write the next message, and after we update our header.
+            // If the application crashes between truncating this message data and writing the next message, our file will still be consistent.
+            if let truncateAtOffset = truncateAtOffset {
+                try writer.truncate(at: truncateAtOffset)
+            }
```
bachand, line 151: I'm having trouble reminding myself why we need to truncate the end of the file. What would really help me in reading this code is a reminder of how the file is formatted. Do we have any graphic or ASCII art anywhere that shows what the format of the file is? If not, I think that would be really helpful to add as it would allow me to validate this code more confidently.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -60,11 +60,17 @@ final class CacheReader {
             return message
 
         case .emptyRead:
+            guard !previousReadWasEmpty else {
+                // If the previous read was also empty, then the file has been corrupted.
+                // Two empty reads in a row means that offsetInFileAtEndOfNewestMessage is incorrect.
+                // This inconsistency likely is likely due to a crash occurring during a message write.
+                throw CacheAdvanceError.fileCorrupted
+            }
```
bachand, line 68: To make sure I follow: we never expect to be in this state anymore, right? It could be useful to make clear that this is a safety net.

bachand (commented):
Thanks for addressing this issue @dfed . I've left a few comments. If you can help refresh my memory on the format of the underlying file (see one of my comments) I plan to revisit this PR. At that point I will be able to review more confidently.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: `FileHeader.expectedEndOfHeaderInFile + 1` does indeed cause the issue, though I don't think that's what the file would have looked like when we were in the bad state. My guess is that we had at least one span in the expected end of newest message, plus the length of that message.

I'm happy to change this to just be `+ 1` since that does reproduce the issue, and since I think we've solve the original issue. I'll do that in a follow-up PR.

Sources/CacheAdvance/CacheAdvance.swift
dfed: We handle that in the following code just above the code you highlighted here:
```
            // Prepare the reader before writing the message.
            try prepareReaderForWriting(dataOfLength: bytesNeededToStoreMessage)
```

This code does the following:
```
    /// Advances the reader until there is room to store a new message without writing past the reader.
    /// This method should only be called on a cache that overwrites old messages.
    /// - Parameter messageLength: the length of the next message that will be written.
    private func prepareReaderForWriting(dataOfLength messageLength: Bytes) throws {
        // If our writer is behind our reader,
        while writer.offsetInFile < reader.offsetInFile
            // And our writer doesn't have enough room to write a message such that it stays behind the current reader position.
            && writer.offsetInFile + messageLength >= reader.offsetInFile
        {
            // Then writing this message would write into the oldest-known message.
            // We must advance our reader to the next-oldest message to help make room for the next message we want to write.
            try reader.seekToNextMessage()
        }
    }
```

We then are willing to write up to where the reader is located after this method returns.

Sources/CacheAdvance/CacheAdvance.swift
dfed: We do not have any ascii art, though I agree that would help a ton.

High level, the format is:
```
[reserved header space]([length-of-message][message-content])*
```

where `([length-of-message][message-content])` can be repeated until the maximum byte count is reached.

Truncating the file prevents us from reading data that should have been deleted when we try to read all messages. Since the only data we have about ordering is the start of first message and end of last message in the header, we'll just try to read all data between these points. If we haven't deleted the data at the end of the file, we would read these supposedly-deleted messages in when trying to read all messages.

Sources/CacheAdvance/CacheReader.swift
dfed: yeah let me update this comment 👍

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,6 +96,35 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
+  func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
```
bachand, line 99: Nit: the indentation on this method is off.

bachand, line 99: Nit: the indentation on this method is off.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -121,6 +144,12 @@ public final class CacheAdvance<T: Codable> {
...
+
```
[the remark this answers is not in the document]
dfed: yes your understanding is correct.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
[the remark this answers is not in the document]
dfed: > The only way this could happen is if the writer (and offset of the start of the oldest message) was at the end of the file's header

I agree with this analysis!

> If we were in this situation, `cacheHasSpaceForNewMessageBeforeEndOfFile` could never be `false`. If `cacheHasSpaceForNewMessageBeforeEndOfFile` were false we would have already failed this previous check, since in this scenario `writer.offsetInFile` would equal `FileHeader.expectedEndOfHeaderInFile`

I disagree!

If we were writing a message of length `header.maximumBytes - FileHeader.expectedEndOfHeaderInFile`, then  `bytesNeededToStoreMessage <= header.maximumBytes - FileHeader.expectedEndOfHeaderInFile` would evaluate as `true` since `bytesNeededToStoreMessage` is equal to `header.maximumBytes - FileHeader.expectedEndOfHeaderInFile`.

If `writer.offsetInFile` is at the `FileHeader.expectedEndOfHeaderInFile` (which I agree is a precondition for this bug), then in the next line `cacheHasSpaceForNewMessageBeforeEndOfFile` would evaluate to `true` because `writer.offsetInFile + bytesNeededToStoreMessage` equals `header.maximumBytes`:
https://github.com/dfed/CacheAdvance/blob/ed9ece4a186b0073514128a75f1cc401ba4ac130/Sources/CacheAdvance/CacheAdvance.swift#L113

At which point, I think we've proved that our original analysis was correct. Am I missing something?

To dig in a bit further, I believe this issue occurred because:
1. `offsetInFileAtEndOfNewestMessage` was beyond the actual end of the file
2. The file was empty except for the header

We only update `offsetInFileAtEndOfNewestMessage` after writing the message: https://github.com/dfed/CacheAdvance/blob/ed9ece4a186b0073514128a75f1cc401ba4ac130/Sources/CacheAdvance/CacheAdvance.swift#L238-L246

Which means that the only way for `offsetInFileAtEndOfNewestMessage` to be beyond the actual end of the file is if we deleted part of the file before updating the header with our next write.

Now, this analysis (thank you for making me write this out!) indicates that my fix to prevent this data corruption in the future was incorrect. The issue isn't when we truncate the file, but rather when we update `offsetInFileAtEndOfNewestMessage`. PR incoming!

dfed: I'll update this in my next PR

dfed: https://github.com/dfed/CacheAdvance/pull/58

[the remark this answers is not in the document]
dfed: https://github.com/dfed/CacheAdvance/pull/58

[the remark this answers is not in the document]
bachand: This explanation makes sense. It seems like I was off by one. Thank you very much for writing this out.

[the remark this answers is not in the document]
bachand: @dfed I made a correction to my original comment. Can you verify that my correction is in fact correct? I am trying to make sure I have the right mental model.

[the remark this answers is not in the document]
bachand: > > If we were in this situation, `cacheHasSpaceForNewMessageBeforeEndOfFile` could never be `false`. If `cacheHasSpaceForNewMessageBeforeEndOfFile` were false we would have already failed this previous check, since in this scenario `writer.offsetInFile` would equal `FileHeader.expectedEndOfHeaderInFile`

> I disagree!

> If we were writing a message of length `header.maximumBytes - FileHeader.expectedEndOfHeaderInFile`, then  `bytesNeededToStoreMessage <= header.maximumBytes - FileHeader.expectedEndOfHeaderInFile` would evaluate as `true` since `bytesNeededToStoreMessage` is equal to `header.maximumBytes - FileHeader.expectedEndOfHeaderInFile`.

> If `writer.offsetInFile` is at the `FileHeader.expectedEndOfHeaderInFile` (which I agree is a precondition for this bug), then in the next line `cacheHasSpaceForNewMessageBeforeEndOfFile` would evaluate to `true` because `writer.offsetInFile + bytesNeededToStoreMessage` equals `header.maximumBytes`:
https://github.com/dfed/CacheAdvance/blob/ed9ece4a186b0073514128a75f1cc401ba4ac130/Sources/CacheAdvance/CacheAdvance.swift#L113

I thought more about this and I've circled back to my original conclusion. Yes, I agree that `cacheHasSpaceForNewMessageBeforeEndOfFile` would evaluate to `true`. However, we only truncate the file if `cacheHasSpaceForNewMessageBeforeEndOfFile` is `false`.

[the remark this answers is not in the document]
bachand: We talked on Zoom and agreed with where I landed in my last comment.

PR #53, opened 2022-03-07
Start using Github Actions

### LINKED ISSUE #51: Move off of Travis CI
Travis CI is sunsetting their program offering free Mac builds for open source projects. To keep CI working we'll need to move to another provider.

dfed:
Resolves #51. This migration was heavily inspired by https://github.com/square/Valet/pull/256

All the new jobs have been listed as required, and the Travis CI job has been removed entirely:
image


Scripts/build.swift
```diff
@@ -23,49 +23,61 @@ enum TaskError: Error {
...
     case macOS_10_15
     case watchOS_5
     case watchOS_6
+    case watchOS_7
 
     var destination: String {
         switch self {
         case .iOS_12:
-            return "platform=iOS Simulator,OS=12.2,name=iPad Pro (12.9-inch) (3rd generation)"
+            return "platform=iOS Simulator,OS=12.4,name=iPad Pro (12.9-inch) (3rd generation)"
```
dfed, line 38: GitHub actions has different simulators pre-installed than Travis did, so we update them as part of this PR

Scripts/build.swift
```diff
@@ -23,49 +23,61 @@ enum TaskError: Error {
 enum Platform: String, CaseIterable, CustomStringConvertible {
     case iOS_12
     case iOS_13
+    case iOS_14
```
dfed, line 26: while I was here I added support for testing on iOS 14. I should add iOS 15 too but... for another day

Scripts/build.swift
```diff
@@ -102,6 +117,12 @@ guard CommandLine.arguments.count > 1 else {
 
 try execute(commandPath: "/usr/bin/swift", arguments: ["package", "generate-xcodeproj", "--output=generated/"])
 
+// The generate-xcodeproj command has a bug where the test deployment target is above the minimum deployment target for the project. Fix it with sed.
```
dfed, line 120: this is absurd, but it works.

Scripts/github/prepare-simulators.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -l
```
dfed, line 1: This is _very_ similar to a file with the same name in Valet. Thank you @NickEntin for figuring out this absurdity.

Scripts/build.swift
```diff
@@ -74,24 +86,27 @@ enum Platform: String, CaseIterable, CustomStringConvertible {
...
+             .watchOS_6,
+             .watchOS_7:
             // watchOS does not support unit testing (yet?).
             return false
         }
     }
 
     var derivedDataPath: String {
-        ".build/derivedData/" + description
+        return ".build/derivedData/" + description
```
dfed, line 105: In order to get access to watchOS 5 on GitHub Actions I needed to run this code on Xcode 10 which means I need `return` statements.

.travis.yml
```diff
@@ -1,42 +0,0 @@
...
-      - bash <(curl -s https://codecov.io/bash) -J '^CADCacheAdvance$' -D .build/derivedData/tvOS_13 -t 8344b011-6b2a-4b3d-a573-eaf49684318e
-
-  - osx_image: xcode11.2
-    env: ACTION="swift-package";PLATFORMS="macOS_10_15";
-    after_success:
-      - bash <(curl -s https://codecov.io/bash) -J '^CacheAdvance$' -D .build/derivedData/macOS_10_15 -t 8344b011-6b2a-4b3d-a573-eaf49684318e
-      - bash <(curl -s https://codecov.io/bash) -J '^CADCacheAdvance$' -D .build/derivedData/macOS_10_15 -t 8344b011-6b2a-4b3d-a573-eaf49684318e
-
-  - osx_image: xcode11.2
-    env: ACTION="swift-package";PLATFORMS="watchOS_5,watchOS_6";
```
dfed, line 32: note that we are no longer testing on watchOS 5. Github Actions requires support for Xcode 10 in order to get access to a watchOS 5 simulator. Not worth the effort to support that old an Xcode version when we've never seen a failure specific to watchOS, let alone watchOS 5.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+  push:
+    branches:
+    - main
+  pull_request:
+
+jobs:
+  pod-lint:
+    name: Pod Lint
+    runs-on: macOS-10.15
+    steps:
```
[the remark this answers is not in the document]
fdiaz: I actually went ahead and took a look, it's possible to have dependent jobs which I think would be the best of both worlds https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#creating-dependent-jobs

bachand:
I will defer to @fdiaz's review. Thanks for fixing completing this migration @dfed .

.github/workflows/ci.yml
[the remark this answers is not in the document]
dfed: Doesn't look like the `needs` API makes the dependent jobs run on the same machine per https://github.com/dfed/CacheAdvance/pull/55 😞

Scripts/github/prepare-simulators.sh
NickEntin: Haha these two lines took me so long to figure out:

```
.../Xcode_11.7.app/.../iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime ...
.../Xcode_10.3.app/.../iPhoneOS.platform/Developer/Library/CoreSimulator/Profiles/Runtimes/iOS.simruntime ...
```

Because, you know, why not mix things up

PR #58, opened 2022-03-10
Limit possibility of file corruption

dfed:
@bachand's [comment](https://github.com/dfed/CacheAdvance/pull/52/files#r822973371) led me to realize that the fix in #52 didn't actually prevent file corruption from happening. Moreover, it is not possible to prevent file corruption entirely!

This PR attempts to minimize the likelihood that we will encounter two `emptyRead`s in a row and throw a file corruption error here:
https://github.com/dfed/CacheAdvance/blob/ed9ece4a186b0073514128a75f1cc401ba4ac130/Sources/CacheAdvance/CacheReader.swift#L63-L69

To repeat some of what [I wrote in response to Michael's review](https://github.com/dfed/CacheAdvance/pull/52/files#r823198935), we can encounter two `emptyRead`s in a row when the:
1. `offsetInFileAtEndOfNewestMessage` is beyond the actual end of the file
2. The file is empty except for the header

We know that in v1.2.0, it is possible to encounter two `emptyRead`s in a row because we saw it happen within the Airbnb app.

We know that the only way for `offsetInFileAtEndOfNewestMessage` to be beyond the actual end of the file is if we deleted part of the file before updating the header with our next write. This means that my root cause fix in #52 wasn't correct. In #52 I attempted to delete data from the file after updating the header's `updateOffsetInFileOfOldestMessage`, when in fact the issue is that we need to delete this data after updating the header's `offsetInFileAtEndOfNewestMessage`.

While I was here I fixed [an indentation issue](https://github.com/dfed/CacheAdvance/pull/52/files#r822993655). **I recommend reviewing this PR commit-by-commit**.

### DIFF
--- a/Sources/CacheAdvance/CacheAdvance.swift
+++ b/Sources/CacheAdvance/CacheAdvance.swift
@@ -112,16 +112,18 @@ public final class CacheAdvance<T: Codable> {
 
         let cacheHasSpaceForNewMessageBeforeEndOfFile = writer.offsetInFile + bytesNeededToStoreMessage <= header.maximumBytes
         if header.overwritesOldMessages {
-            let truncateAtOffset: UInt64?
-            if cacheHasSpaceForNewMessageBeforeEndOfFile {
-                // We have room for this message. No need to truncate.
-                truncateAtOffset = nil
-            } else {
+            if !cacheHasSpaceForNewMessageBeforeEndOfFile {
                 // This message can't be written without exceeding our maximum file length.
                 // We'll need to start writing the file from the beginning of the file.
 
                 // Trim the file to the current writer position to remove soon-to-be-abandoned data from the file.
-                truncateAtOffset = writer.offsetInFile
+                try writer.truncate(at: writer.offsetInFile)
+
+                // Update the offsetInFileAtEndOfNewestMessage in our header and reader such that we will ignore the now-deleted data in the file.
+                // If the application crashes between writing this header and writing the next message data, we'll have only lost the messages we were about to delete.
+                // If `writer.offsetInFile == FileHeader.expectedEndOfHeaderInFile`, then crashing before we update the header would lead to this now-empty file being viewed as corrupted.
+                try header.updateOffsetInFileAtEndOfNewestMessage(to: writer.offsetInFile)
+                reader.offsetInFileAtEndOfNewestMessage = writer.offsetInFile
 
                 // Set the offset back to the beginning of the file.
                 try writer.seek(to: FileHeader.expectedEndOfHeaderInFile)
@@ -143,12 +145,6 @@ public final class CacheAdvance<T: Codable> {
             // Update the offsetInFileOfOldestMessage in our header before we write the message.
             // If the application crashes between writing the header and writing the message data, we'll have lost the messages between the previous offsetInFileOfOldestMessage and the new offsetInFileOfOldestMessage.
             try header.updateOffsetInFileOfOldestMessage(to: offsetInFileOfOldestMessage)
-
-            // Truncate the file if it needs truncation before we write the next message, and after we update our header.
-            // If the application crashes between truncating this message data and writing the next message, our file will still be consistent.
-            if let truncateAtOffset = truncateAtOffset {
-                try writer.truncate(at: truncateAtOffset)
-            }
 
             // Let the reader know where the oldest message begins.
             reader.offsetInFileOfOldestMessage = offsetInFileOfOldestMessage
--- a/Sources/CacheAdvance/CacheReader.swift
+++ b/Sources/CacheAdvance/CacheReader.swift
@@ -63,8 +63,8 @@ final class CacheReader {
             guard !previousReadWasEmpty else {
                 // If the previous read was also empty, then the file has been corrupted.
                 // Two empty reads in a row means that offsetInFileAtEndOfNewestMessage is incorrect.
-                // This inconsistency is likely due to a crash occurring during a message write.
-                // This issue will not occur in a current version of the library, but an earlier version was capable of creating this corruption.
+                // This inconsistency is likely due to a crash occurring during a message write, between truncating the file and updating the header values to reflect that the file was truncated.
+                // This file is actually empty, despite what the header tells us.
                 throw CacheAdvanceError.fileCorrupted
             }
             // We know the next message is at the end of the file header. Let's seek to it.
--- a/CacheAdvance.podspec
+++ b/CacheAdvance.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '1.2.1'
+  s.version  = '1.2.2'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'A performant cache for logging systems. CacheAdvance persists log events 30x faster than SQLite.'
   s.homepage = 'https://github.com/dfed/CacheAdvance'
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -96,34 +96,34 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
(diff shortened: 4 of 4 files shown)


Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -112,16 +112,18 @@ public final class CacheAdvance<T: Codable> {
...
                 // This message can't be written without exceeding our maximum file length.
                 // We'll need to start writing the file from the beginning of the file.
 
                 // Trim the file to the current writer position to remove soon-to-be-abandoned data from the file.
-                truncateAtOffset = writer.offsetInFile
+                try writer.truncate(at: writer.offsetInFile)
+
+                // Update the offsetInFileAtEndOfNewestMessage in our header and reader such that we will ignore the now-deleted data in the file.
+                // If the application crashes between writing this header and writing the next message data, we'll have only lost the messages we were about to delete.
+                // If `writer.offsetInFile == FileHeader.expectedEndOfHeaderInFile`, then crashing before we update the header would lead to this now-empty file being viewed as corrupted.
```
dfed, line 124: I couldn't think of a way to eliminate this issue. If we truncate after writing the header, then we can still end up with file corruption because our `reader` doesn't actually know when the file _should_ stop before it reaches `maximumBytes`. So if we had:
```
[header][length|long-message7][length|message5][length|message6]garbage-remnant-of-message-3
                                                                ^ reader                      ^ maximumBytes
```

our `nextEncodedMessageSpan()` method would try to read the length of the the garbage remnant of the message three, and we'd end up trying to decode garbage.

Ideally our header would keep track of the last non-garbage byte in the file and we wouldn't need to truncate the file, but that's a much larger lift.

bachand: > Ideally our header would keep track of the last non-garbage byte in the file and we wouldn't need to truncate the file, but that's a much larger lift.

This makes sense though I worry that there's a new chance of getting out of sync. Let's say we started with an empty file and write a message. We'd need to update both the offset of the end of the newest message, AND the offset of the last non-garbage byte in the file. We can't update both of those values atomically. In writing this comment I also realized (or re-realized) that none of our writes are guaranteed to be atomic, which is a scary thought.

bachand (approved):

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -63,8 +63,8 @@ final class CacheReader {
             guard !previousReadWasEmpty else {
                 // If the previous read was also empty, then the file has been corrupted.
                 // Two empty reads in a row means that offsetInFileAtEndOfNewestMessage is incorrect.
-                // This inconsistency is likely due to a crash occurring during a message write.
-                // This issue will not occur in a current version of the library, but an earlier version was capable of creating this corruption.
+                // This inconsistency is likely due to a crash occurring during a message write, between truncating the file and updating the header values to reflect that the file was truncated.
+                // This file is actually empty, despite what the header tells us.
                 throw CacheAdvanceError.fileCorrupted
```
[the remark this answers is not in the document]
dfed: I think this would be a nice improvement over what we have, but since it's not that trivial to write and we're talking about a very, very rare case I'm not sure it's worth the effort. I do agree though that throwing `fileCorrupted` when we can validate that in fact we really have no messages isn't great. But I think it's also reasonable (though not ideal) to say that the file is corrupted when we know for certain that the header and the file contents are out of sync. And since we are expecting that this case is due to the file being unexpectedly empty, we expect that throwing here doesn't translate to lost data.

Sources/CacheAdvance/CacheAdvance.swift
dfed: Yes the approach of keeping track of the last non-garbage byte in the file does lead to more chances that we get out of sync. However, we could:
1. Write the `header.offsetOfEndOfFile`. If we crash here we've just lost the data we were about to delete. We can accomplish this by making our reading code trust `offsetOfEndOfFile` more than it trusts `updateOffsetInFileAtEndOfNewestMessage` (i.e. `if offsetOfEndOfFile < updateOffsetInFileAtEndOfNewestMessage  { /* assume updateOffsetInFileAtEndOfNewestMessage is corrupt and trust offsetOfEndOfFile instead */ }`).
2. Write the `header.offsetInFileAtEndOfNewestMessage`. If we crash here we've just lost the data we were about to delete.

Basically I think adding this data to the header is workable, plugs a known hole in our existing logic, and would not lead to more potential file corruption. That said I don't think plugging this hole is worth the additional effort. Right now a crash during a write won't lead to data loss, which feels like the right bar to hit.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -112,16 +112,18 @@ public final class CacheAdvance<T: Codable> {
...
+                try header.updateOffsetInFileAtEndOfNewestMessage(to: writer.offsetInFile)
```
bachand, line 125: I want to dig into this a bit more... Isn't the writer _always_ at the end of the newest message? Under what circumstances would this line of code have an effect?

bachand (commented):
I had another comment I'm chewing on 🤔

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -144,12 +146,6 @@ public final class CacheAdvance<T: Codable> {
             // If the application crashes between writing the header and writing the message data, we'll have lost the messages between the previous offsetInFileOfOldestMessage and the new offsetInFileOfOldestMessage.
             try header.updateOffsetInFileOfOldestMessage(to: offsetInFileOfOldestMessage)
 
-            // Truncate the file if it needs truncation before we write the next message, and after we update our header.
```
bachand, line 147: Thinking about this more I'm not sure there is a difference from a data safety POV in truncating the file before or after we update the offset of the start of the oldest message:

1. If the offset of the start of the oldest message is before the offset of the end of the newest message there's no way we can ever read into the bytes after the end of the newest message.
2. If the offset of the start of the oldest message and the offset of the end of the newest message are both at the end of the header we will never truncate (see https://github.com/dfed/CacheAdvance/pull/52#discussion_r822973371).
3. We only need to consider the case where the offset of the start of the oldest message is after the offset of the end of the newest message.

For (3) if we truncate immediately (as we are changing this code to do in this PR), we could crash after truncation and before we update the offset of the start of the oldest message. In this situation the offset of the start of the oldest message is in the truncation region. When the file was read again we'd have an empty read and go back to the beginning of the file.

For (3) if we truncate after updating the offset of the start of the oldest message we could crash before truncating. In this situation we'd have non-0 bytes at the end of the file. But the reader will never get to these bytes since they are after the offset of the end of the newest message, like in the first situation.

@dfed do you agree with this reasoning? I'm OK with truncating earlier but I wanted to make sure I understand.

Sources/CacheAdvance/CacheAdvance.swift
dfed: good point! this line is indeed duplicative.

so... now I'm really getting lost on how our file got into the corrupted state in #52. It's almost as if we wrote an empty message to the file... but that shouldn't be possible?!

Sources/CacheAdvance/CacheAdvance.swift
dfed: > For (3) if we truncate immediately (as we are changing this code to do in this PR), we could crash after truncation and before we update the offset of the start of the oldest message. In this situation the offset of the start of the oldest message is in the truncation region. When the file was read again we'd have an empty read and go back to the beginning of the file.

This is my understanding.

> For (3) if we truncate after updating the offset of the start of the oldest message we could crash before truncating. In this situation we'd have non-0 bytes at the end of the file. But the reader will never get to these bytes since they are after the offset of the end of the newest message, like in the first situation.

Agreed.

> do you agree with this reasoning?

Yes. And I'm coming to realize that I don't think that #52 or this PR actually reduced (or increased) the likelihood of a file corruption. #52 _did_ help handle the file corruption in a better wya than we were doing before. But I no longer think I understand how we ended up with the file corruption, which isn't great.

Sources/CacheAdvance/CacheReader.swift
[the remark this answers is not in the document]
dfed: Per [discussion elsewhere on this PR](https://github.com/dfed/CacheAdvance/pull/58/files#r824351117) I am no longer convinced this comment is correct.

PR #60, opened 2022-03-14
Remove build tag from the readme

dfed:
Now that we're off TravisCI, we should remove our README badge that points at travis


README.md
```diff
@@ -1,6 +1,5 @@
 # CacheAdvance
 
-[![CI Status](https://travis-ci.com/dfed/CacheAdvance.svg?branch=main)](https://travis-ci.com/dfed/CacheAdvance)
```
dfed, line 3: it's possible that we can get this working with GitHub Actions but I'm not convinced this badge has ever been super helpful

bachand (approved):

PR #62, opened 2022-09-02
Update Bundle and CocoaPods

dfed:
This PR accomplishes three goals

### Adopt Modern Gemfile Standards
This change gets our Gemfile to use current standards by adopting a global (rather than scoped) `source`. Prior to this change, we'd see the following warning when running `bundle install`:
> [DEPRECATED] This Gemfile does not include an explicit global source. Not using an explicit global source may result in a different lockfile being generated depending on the gems you have installed locally before bundler is run. Instead, define a global source in your Gemfile like this: source "https://rubygems.org".

### Updating our bundle
This change is the result of running `bundle update`. It should fix [this dependabot issue](https://github.com/dfed/CacheAdvance/security/dependabot/3).

### Updating CocoaPods
We were using version 1.8. Now we're using 1.11


fdiaz (approved):

dfed:
@fdiaz can you give this another look?

PR #63, opened 2022-09-02
Update CI to run on macOS 11

### LINKED ISSUE #44: Build and test in CI using `swift` (in addition to `xcodebuild`)
Inspired by https://github.com/dfed/CacheAdvance/pull/42#discussion_r415180088.

**Goals**
- Ensure that changes to this library don't break building and testing with `swift` directly.

**Non-Goals**
- Test all OS configurations using `swift`. We can continue to have our comprehensive test matrix use `xcodebuild`.
- Remove tests that rely on `xcodebuild` in favor of `swift`. ROI is not there, and its nice to ensure that the generated project works.

**Investigations**
n/a

**Design**

Build and test via `swift` in CI to sanity to check that this continues to work.

dfed:
macOS 10.15 [is no longer supported by Github Actions](https://github.com/actions/runner-images/blob/main/images/macos/macos-10.15-Readme.md). So instead let's run CI on macOS 11 and macOS 12.

While I was here, I also added Xcode 13 CI runs, and resolved #44

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,4 +116,4 @@ jobs:
       - name: Select Xcode Version
         run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
       - name: Build and Test Framework
-        run: swift test -c release -Xswiftc -enable-testing
+        run: xcrun swift test -c release -Xswiftc -enable-testing
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,14 +34,14 @@ jobs:
       - name: Build Framework
         run: swift package generate-xcodeproj --output generated/ && carthage build --project-directory generated/ --verbose --no-skip-current
   spm-11:
-    name: SPM Build Xcode 11
+    name: Build Xcode 11
     runs-on: macOS-11
     strategy:
       matrix:
         platforms: [
           'iOS_12,tvOS_12',
           'iOS_13,tvOS_13,watchOS_6',
-          'macOS_11'
+          'macOS_10_15',
         ]
       fail-fast: false
     steps:
@@ -59,12 +59,13 @@ jobs:
         if: success()
         run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
   spm-12:
-    name: SPM Build Xcode 12
+    name: Build Xcode 12
     runs-on: macOS-11
     strategy:
       matrix:
         platforms: [
           'iOS_14,tvOS_14,watchOS_7',
+          'macOS_11',
         ]
       fail-fast: false
     steps:
@@ -81,3 +82,39 @@ jobs:
       - name: Upload Coverage Reports
         if: success()
         run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
+  spm-13:
+    name: Build Xcode 13
+    runs-on: macOS-12
+    strategy:
+      matrix:
+        platforms: [
+          'iOS_15,tvOS_15,watchOS_8',
+          'macOS_12',
+        ]
+      fail-fast: false
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Bundle Install
+        run: bundle install
+      - name: Select Xcode Version
+        run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
+      - name: Prepare Simulator Runtimes
+        run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
+      - name: Build and Test Framework
+        run: Scripts/build.swift ${{ matrix.platforms }}
+      - name: Upload Coverage Reports
+        if: success()
+        run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
+  spm-13-swift:
+    name: Swift Build Xcode 13
+    runs-on: macOS-12
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Bundle Install
+        run: bundle install
+      - name: Select Xcode Version
+        run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
+      - name: Build and Test Framework
+        run: swift test -c release -Xswiftc -enable-testing
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
+        run: xcversion select 11.7
       - name: Lint Podspec
         run: bundle exec pod lib lint --verbose --fail-fast --swift-version=5.1
   carthage:
@@ -28,7 +28,7 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
+        run: xcversion select 11.7
       - name: Install Carthage
         run: brew outdated carthage || brew upgrade carthage
       - name: Build Framework
@@ -49,7 +49,7 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
+        run: xcversion select 11.7
       - name: Prepare Simulator Runtimes
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
@@ -73,7 +73,7 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_12.4.app/Contents/Developer
+        run: xcversion select 12.4
       - name: Prepare Simulator Runtimes
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
@@ -97,7 +97,7 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
+        run: xcversion select 13.4.1
       - name: Prepare Simulator Runtimes
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
@@ -114,6 +114,6 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
+        run: xcversion select 13.4.1
       - name: Build and Test Framework
         run: xcrun swift test -c release -Xswiftc -enable-testing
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -35,13 +35,13 @@ jobs:
...
     name: SPM Build Xcode 11
-    runs-on: macOS-10.15
+    runs-on: macOS-11
     strategy:
       matrix:
         platforms: [
           'iOS_12,tvOS_12',
           'iOS_13,tvOS_13,watchOS_6',
-          'macOS_10_15'
+          'macOS_11'
```
dfed, line 44: I don't love this change but not much we can do about it. We still technically support back to macOS 10.14, even though we haven't been able to run CI on that OS in years 😬

Scripts/build.swift
```diff
@@ -115,31 +164,43 @@ guard CommandLine.arguments.count > 1 else {
...
-        "OTHER_CFLAGS='-DGENERATED_XCODE_PROJECT'",
+        "-PBXBuildsContinueAfterErrors=0"
     ]
+    if platform.shouldGenerateXcodeproj {
+        // Point at the generated project
+        xcodeBuildArguments.append("-project")
+        xcodeBuildArguments.append("generated/CacheAdvance.xcodeproj")
+        // Set the configuration to be release – this configuration is not supported when running xcodebuild without a xcodeproj file.
+        xcodeBuildArguments.append("-configuration")
+        xcodeBuildArguments.append("Release")
```
dfed, line 200: This configuration only existing when we have an xcodeproj file sucks. I managed to continue testing in Release in CI by using `swift test  -c release -Xswiftc -enable-testing`. We're still running via `xcodebuild` for test coverage on Xcode 13, but that run is in debug rather than Release.

.github/workflows/ci.yml
```diff
@@ -81,3 +82,39 @@ jobs:
...
+    runs-on: macOS-12
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Bundle Install
+        run: bundle install
+      - name: Select Xcode Version
+        run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
+      - name: Build and Test Framework
+        run: swift test -c release -Xswiftc -enable-testing
```
dfed, line 120: These new lines resolve #44

.github/workflows/ci.yml
```diff
@@ -34,14 +34,14 @@ jobs:
...
-    runs-on: macOS-10.15
+    name: Build Xcode 11
+    runs-on: macOS-11
     strategy:
       matrix:
         platforms: [
           'iOS_12,tvOS_12',
           'iOS_13,tvOS_13,watchOS_6',
-          'macOS_10_15'
+          'macOS_10_15',
```
dfed, line 44: Fun fact, we can still run tests against the macOS 10.15 SDK even while running on macOS 11. We can only do this while running on Xcode 11.

.github/workflows/ci.yml
```diff
@@ -59,12 +59,13 @@ jobs:
...
   spm-12:
-    name: SPM Build Xcode 12
-    runs-on: macOS-10.15
+    name: Build Xcode 12
+    runs-on: macOS-11
     strategy:
       matrix:
         platforms: [
           'iOS_14,tvOS_14,watchOS_7',
+          'macOS_11',
```
dfed, line 68: Now that we're running on macOS 11 and targeting Xcode 12 we can run tests against the macOS 11 SDK!

.github/workflows/ci.yml
```diff
@@ -34,14 +34,13 @@ jobs:
...
         run: swift package generate-xcodeproj --output generated/ && carthage build --project-directory generated/ --verbose --no-skip-current
   spm-11:
-    name: SPM Build Xcode 11
-    runs-on: macOS-10.15
+    name: Build Xcode 11
+    runs-on: macOS-11
     strategy:
       matrix:
         platforms: [
-          'iOS_12,tvOS_12',
```
dfed, line 42: This sucks, but [Github Actions' macOS 11 install does not have iOS or tvOS 12 simulators](https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md)

Scripts/build.swift
```diff
@@ -21,82 +21,113 @@ enum TaskError: Error {
...
-    case iOS_12
     case iOS_13
     case iOS_14
-    case tvOS_12
+    case iOS_15
     case tvOS_13
     case tvOS_14
+    case tvOS_15
     case macOS_10_15
-    case watchOS_5
```
dfed, line 31: I don't know why `watchOS_5` was here: we weren't running tests on it!

.github/workflows/ci.yml
```diff
@@ -81,3 +81,39 @@ jobs:
...
+          'macOS_12',
+        ]
+      fail-fast: false
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Bundle Install
+        run: bundle install
+      - name: Select Xcode Version
+        run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
```
fdiaz, line 100: JFYI `xcversion` comes installed in these machines, which would make selecting an Xcode much easier than this https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#xcode-support-tools

Scripts/build.swift
```diff
@@ -87,20 +107,49 @@ enum Platform: String, CaseIterable, CustomStringConvertible {
...
+             .watchOS_5,
+             .watchOS_6,
+             .watchOS_7:
+            return true
+
+        case .iOS_15,
+             .tvOS_15,
+             .macOS_12,
+             .watchOS_8:
+            // Xcode 13 does not require xcodeproj generation
```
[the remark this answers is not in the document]
fdiaz: Interesting!

.github/workflows/ci.yml
fdiaz: Qing has told me to always use `xcrun swift` for these things which IIRC it's so we use the `swift` version from Xcode rather than the one installed in the system.

fdiaz (approved):

.github/workflows/ci.yml
```diff
@@ -81,3 +81,39 @@ jobs:
...
+    runs-on: macOS-12
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Bundle Install
+        run: bundle install
+      - name: Select Xcode Version
+        run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
+      - name: Build and Test Framework
+        run: swift test -c release -Xswiftc -enable-testing
```
dfed, line 119: Let's see if [this](https://github.com/dfed/CacheAdvance/pull/63/files#r961946075) works!
```suggestion
        run: xcrun swift test -c release -Xswiftc -enable-testing
```

Sources/CacheAdvance/BigEndianHostSwappable.swift
```diff
@@ -31,10 +31,10 @@ extension BigEndianHostSwappable {
     ///
     /// - Parameter data: A data blob representing encodable data. Must be of length `Self.storageLength`.
     init(_ data: Data) {
-        let decodedSize = withUnsafePointer(to: data) {
-            return UnsafeRawBufferPointer(start: $0, count: Self.storageLength)
+        let decodedSize = data.withUnsafeBytes {
+            return $0.load(as: Self.self)
         }
-        self = Self.swapToHost(decodedSize.load(as: Self.self))
+        self = Self.swapToHost(decodedSize)
```
dfed, line 37: Woof. I don't understand why this change was necessary to get CI passing in Release on Xcode 13 – Debug worked fine.

The good news is the only tests that were failing were:
```
-[CacheAdvanceTests.BigEndianHostSwappableTests test_init_canBeInitializedFromEncodedData]
-[CacheAdvanceTests.EncodableMessageTests test_encodedData_encodesCorrectSize]
```

These tests are effectively testing implementation details – the integration tests where we persist messages and de-persist them were still succeeding.

So, what does this change do? Instead of creating a buffer pointer from the data's pointer, I instead access the underlying data's bytes directly. Skipping the translation to a buffer pointer simplifies what's happening here, which fixed the underlying issue. Skipping the translation buffer likely also gave us a very small performance win.

.github/workflows/ci.yml
dfed: Nice! de95961

Sources/CacheAdvance/BigEndianHostSwappable.swift
```diff
@@ -31,10 +31,10 @@ extension BigEndianHostSwappable {
     ///
     /// - Parameter data: A data blob representing encodable data. Must be of length `Self.storageLength`.
     init(_ data: Data) {
-        let decodedSize = withUnsafePointer(to: data) {
```
fdiaz, line 34: I don't have enough context to review these changes but everything else LGTM!

fdiaz (approved):

Sources/CacheAdvance/BigEndianHostSwappable.swift
dfed: I am confident in these changes (we have _excellent_ test coverage), so I'm okay merging this in.

What's happening in this method is we're instantiating a `BigEndianHostSwappable` (i.e. a `UInt64`, `UInt32`, or `UInt8`) from a `Data` blob.

The way we did that was:
1. Get a pointer to the underlying data (`withUnsafePointer(to: data) { ... }`)
2. Turn that pointer into a buffer pointer (i.e. a pointer range) starting at the pointer and ending at the length of storage required to store this type (`UnsafeRawBufferPointer(start: $0, count: Self.storageLength)`)
3. Using that buffer pointer to `load` the underlying `BigEndianHostSwappable`, i.e. read the pointer buffer from memory and cast it as the current type. (`decodedSize.load(as: Self.self)`)
4. Swap the read-in value to have the same Endian-ness as the current machine – this is only necessary if the file has been transferred between machines of different Endian-ness. (`Self.swapToHost(...)`)

The way we're doing it now is:
1. Get the underlying buffer pointer for the underlying data (`withUnsafeBytes { ... }`)
2. Using that buffer pointer (i.e. a pointer range) to `load` the underlying `BigEndianHostSwappable`, i.e. read the pointer buffer from memory and cast it as the current type. (`$0.load(as: Self.self)`)
3. Swap the read-in value to have the same Endian-ness as the current machine – this is only necessary if the file has been transferred between machines of different Endian-ness. (`Self.swapToHost(...)`)

The issue with the old code (which of course I figured out by typing all of this out) was that we were working over the unsafe buffer pointer range _outside_ of the `withUnsafePointer` call. If I change this code to read:
```
        let decodedSize = withUnsafePointer(to: data) {
            return UnsafeRawBufferPointer(start: $0, count: Self.storageLength).load(as: Self.self)
        }
```
The tests pass in Release! So changing from `withUnsafePointer` to `withUnsafeBytes` wasn't the issue. Instead, it was where we were interacting with the buffer! That said, I do like the semantics of `withUnsafeBytes` better.

dfed:
I have made all the new CI checks required. Merging!

Sources/CacheAdvance/BigEndianHostSwappable.swift
bachand: To make sure I follow, the issue was that we were violating this requirement?
Screen Shot 2022-09-02 at 6 47 19 PM

bachand (commented):
I reviewed the changes to BigEndianHostSwappable. Thanks for the explanation Dan!

Sources/CacheAdvance/BigEndianHostSwappable.swift
dfed: Exactly!

PR #59, opened 2022-03-11
Clean up after #52 and #58

dfed:
Wrapping up some good discussions from #52 and #58

PR #64, opened 2022-10-09
Add readerForwardOnly to CacheAdvance init parameter to avoid possible infinite loop

jianjunwoo:
Hi @dfed , did you remember the JSON decoding optimization in Jitney? There was OOM increase in the treatment group of Jitney uploader optimization. Finally and hopefully, I fix it this time. Very glad to share this with you, and please take a look at this PR.

This is a brief how this problem got figure out.  I added breadcrumbs before and after `func messages()`.  According to breadcrumbs on bugsnag and Xcode energy reports, I have a hypothesis that there is a possible infinite loop when call `func messages()`. 

Then I investigated the code, and found there is potential infinite loop case when the offset in the header is not correct. Please see the details in the unit tests of the PR.

This recording shows the infinite loop triggered in unit test with one message written:

infinite_loop

This fix is really last for a long time for me working on this. Hope this times works : )

### DIFF
--- a/Sources/CacheAdvance/CacheReader.swift
+++ b/Sources/CacheAdvance/CacheReader.swift
@@ -65,10 +65,12 @@ final class CacheReader {
                 // If the previous read was also empty, then the file has been corrupted.
                 throw CacheAdvanceError.fileCorrupted
             }
-            if seekForwardOnly && startingOffset > FileHeader.expectedEndOfHeaderInFile {
-                // If seekForwardOnly is true, then return nil when startingOffset is ahead
-                // Otherwise, there will be an infinite loop in reading messages.
-                return nil
+            if seekForwardOnly {
+                guard startingOffset <= FileHeader.expectedEndOfHeaderInFile else {
+                    // If the startingOffset is ahead of `expectedEndOfHeaderInFile`, return nil to stop reading.
+                    // Otherwise, there will be an infinite loop in reading messages.
+                    return nil
+                }
             }
             // We know the next message is at the end of the file header. Let's seek to it.
             try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)
--- a/Sources/CacheAdvance/CacheReader.swift
+++ b/Sources/CacheAdvance/CacheReader.swift
@@ -24,8 +24,9 @@ final class CacheReader {
     /// Creates a new instance of the receiver.
     ///
     /// - Parameter file: The file URL indicating the desired location of the on-disk store. This file should already exist.
-    init(forReadingFrom file: URL) throws {
+    init(forReadingFrom file: URL, seekForwardOnly: Bool = false) throws {
         reader = try FileHandle(forReadingFrom: file)
+        self.seekForwardOnly = seekForwardOnly
     }
 
     deinit {
@@ -63,6 +64,11 @@ final class CacheReader {
             guard !previousReadWasEmpty else {
                 // If the previous read was also empty, then the file has been corrupted.
                 throw CacheAdvanceError.fileCorrupted
+            }
+            if seekForwardOnly && startingOffset > FileHeader.expectedEndOfHeaderInFile {
+                // If seekForwardOnly is true, then return nil when startingOffset is ahead
+                // Otherwise, there will be an infinite loop in reading messages.
+                return nil
             }
             // We know the next message is at the end of the file header. Let's seek to it.
             try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)
@@ -116,6 +122,7 @@ final class CacheReader {
     }
 
     private let reader: FileHandle
+    private let seekForwardOnly: Bool
 
 }
 
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -97,65 +97,18 @@ final class CacheAdvanceTests: XCTestCase {
     }
 
     func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
-        let randomHighValue: UInt64 = 10_1000
-        let header = try CacheHeaderHandle(
-            forReadingFrom: testFileLocation,
-            maximumBytes: randomHighValue,
-            overwritesOldMessages: true)
-        let cache = CacheAdvance<TestableMessage>(
-            fileURL: testFileLocation,
-            writer: try FileHandle(forWritingTo: testFileLocation),
-            reader: try CacheReader(forReadingFrom: testFileLocation),
-            header: try CacheHeaderHandle(
-                forReadingFrom: testFileLocation,
-                maximumBytes: header.maximumBytes,
-                overwritesOldMessages: header.overwritesOldMessages),
-            decoder: JSONDecoder(),
-            encoder: JSONEncoder())
-
-        // Make sure the header data is persisted before we read it as part of the `messages()` call below.
-        try header.synchronizeHeaderData()
-        // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
-        // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
-        try header.updateOffsetInFileAtEndOfNewestMessage(
-            to: FileHeader.expectedEndOfHeaderInFile + 1)
-
+        let header = try prepareIncorrectOffsetInHeaderAndWriteMessage(messageCount: 0)
+        let cache = try CacheAdvance<TestableMessage>(fileURL: testFileLocation,
+                                                      maximumBytes: header.maximumBytes,
+                                                      shouldOverwriteOldMessages: header.overwritesOldMessages)
         XCTAssertThrowsError(try cache.messages()) {
             XCTAssertEqual($0 as? CacheAdvanceError, CacheAdvanceError.fileCorrupted)
         }
     }
 
-    private func prepareCacheWithIncorrectOffsetInHeader() throws -> CacheHeaderHandle {
-        let randomHighValue: UInt64 = 10_1000
-        let header = try CacheHeaderHandle(
-            forReadingFrom: testFileLocation,
-            maximumBytes: randomHighValue,
-            overwritesOldMessages: true)
-        let cacheToWrite = CacheAdvance<TestableMessage>(
-            fileURL: testFileLocation,
-            writer: try FileHandle(forWritingTo: testFileLocation),
-            reader: try CacheReader(forReadingFrom: testFileLocation),
-            header: try CacheHeaderHandle(
-                forReadingFrom: testFileLocation,
-                maximumBytes: header.maximumBytes,
-                overwritesOldMessages: header.overwritesOldMessages),
-            decoder: JSONDecoder(),
-            encoder: JSONEncoder())
-
-        let message = TestableMessage(stringLiteral: TestableMessage.lorumIpsum[0].value)
-        try cacheToWrite.append(message: message)
-
-        // Make sure the header data is persisted before we read it as part of the `messages()` call below.
-        try header.synchronizeHeaderData()
-        // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
-        // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
-        try header.updateOffsetInFileAtEndOfNewestMessage(to: header.offsetInFileAtEndOfNewestMessage + 1000)
-        return header
-    }
-
     func test_messages_timeout_WhenOffsetInFileAtEndOfNewestMessageLargerThanActualFileLength() throws {
         do {
-            let header = try prepareCacheWithIncorrectOffsetInHeader()
+            let header = try prepareIncorrectOffsetInHeaderAndWriteMessage(messageCount: 1)
             let cacheToRead = try CacheAdvance<TestableMessage>(fileURL: testFileLocation,
                                                                 maximumBytes: header.maximumBytes,
                                                                 shouldOverwriteOldMessages: header.overwritesOldMessages)
@@ -180,7 +133,7 @@ final class CacheAdvanceTests: XCTestCase {
 
     func test_messages_seekForwardOnlyTrue_WhenOffsetInFileAtEndOfNewestMessageLargerThanActualFileLength() throws {
         do {
-            let header = try prepareCacheWithIncorrectOffsetInHeader()
+            let header = try prepareIncorrectOffsetInHeaderAndWriteMessage(messageCount: 1)
             let cacheToRead = try CacheAdvance<TestableMessage>(fileURL: testFileLocation,
                                                                 maximumBytes: header.maximumBytes,
                                                                 shouldOverwriteOldMessages: header.overwritesOldMessages,
@@ -724,6 +677,39 @@ final class CacheAdvanceTests: XCTestCase {
             atPath: testFileLocation.path,
             contents: nil,
             attributes: nil)
+    }
+
+    /// Creates a cache file with incorrect offset in header.
+    /// - Parameter messageCount: The number of messages to write to the cache file
+    /// Returns the header handle of the cache.
+    private func prepareIncorrectOffsetInHeaderAndWriteMessage(messageCount: Int) throws -> CacheHeaderHandle {
+        let randomHighValue: UInt64 = 100_000
+        let header = try CacheHeaderHandle(
+            forReadingFrom: testFileLocation,
+            maximumBytes: randomHighValue,
+            overwritesOldMessages: true)
+        let cacheToWrite = CacheAdvance<TestableMessage>(
+            fileURL: testFileLocation,
+            writer: try FileHandle(forWritingTo: testFileLocation),
+            reader: try CacheReader(forReadingFrom: testFileLocation),
+            header: try CacheHeaderHandle(
+                forReadingFrom: testFileLocation,
+                maximumBytes: header.maximumBytes,
+                overwritesOldMessages: header.overwritesOldMessages),
+            decoder: JSONDecoder(),
+            encoder: JSONEncoder())
+
+        for _ in 0..<messageCount {
+            let message = TestableMessage(stringLiteral: TestableMessage.lorumIpsum[0].value)
+            try cacheToWrite.append(message: message)
+        }
+
+        // Make sure the header data is persisted before we read it as part of the `messages()` call below.
+        try header.synchronizeHeaderData()
+        // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
+        // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
+        try header.updateOffsetInFileAtEndOfNewestMessage(to: header.offsetInFileAtEndOfNewestMessage + 1)
+        return header
     }
 
     private let testFileLocation = FileManager.default.temporaryDirectory.appendingPathComponent("CacheAdvanceTests")
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -151,7 +151,6 @@ final class CacheAdvanceTests: XCTestCase {
         }
     }
 
-
     func test_isWritable_returnsTrueWhenStaticHeaderMetadataMatches() throws {
         let originalCache = try createCache(overwritesOldMessages: false)
         XCTAssertTrue(try originalCache.isWritable())
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -134,6 +134,36 @@ final class CacheAdvanceTests: XCTestCase {
         wait(for: [expectation], timeout: 2)
     }
 
+  func test_messages_timeout_WhenOffsetInFileAtEndOfNewestMessageSmallThanActualFileLength() throws {
+      do {
+          let message = TestableMessage(stringLiteral: TestableMessage.lorumIpsum[0].value)
+          let header = try prepareIncorrectOffsetInHeaderAndWriteMessage(messages: [message],
+                                                                         randomHighValue: 103_000,
+                                                                         offsetInFileAtEndOfNewestMessageToAdd: 0,
+                                                                         offsetInFileAtEndOfNewestMessageToMinus: 1)
+          let cacheToRead = try CacheAdvance<TestableMessage>(fileURL: testFileLocation,
+                                                              maximumBytes: header.maximumBytes,
+                                                              shouldOverwriteOldMessages: header.overwritesOldMessages)
+          DispatchQueue.global().async {
+              do {
+                  let _ = try cacheToRead.messages()
+                  // messages() should be in an infinite loop, and never return
+                  XCTFail()
+              } catch {
+                  XCTFail()
+              }
+          }
+      } catch {
+          XCTFail()
+      }
+      let expectation = XCTestExpectation()
+      DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
+          expectation.fulfill()
+      }
+      wait(for: [expectation], timeout: 2)
+  }
+
+
     func test_messages_seekForwardOnlyTrue_WhenOffsetInFileAtEndOfNewestMessageLargerThanActualFileLength() throws {
         do {
             let message = TestableMessage(stringLiteral: TestableMessage.lorumIpsum[0].value)
@@ -687,7 +717,9 @@ final class CacheAdvanceTests: XCTestCase {
     /// - Parameter messageCount: The number of messages to write to the cache file
     /// Returns the header handle of the cache.
     private func prepareIncorrectOffsetInHeaderAndWriteMessage(messages: [TestableMessage],
-                                                               randomHighValue: UInt64 = 101_000) throws -> CacheHeaderHandle {
+                                                               randomHighValue: UInt64 = 101_000,
+                                                               offsetInFileAtEndOfNewestMessageToAdd: UInt64 = 1,
+                                                               offsetInFileAtEndOfNewestMessageToMinus: UInt64 = 0) throws -> CacheHeaderHandle {
         let header = try CacheHeaderHandle(
             forReadingFrom: testFileLocation,
             maximumBytes: randomHighValue,
@@ -709,7 +741,10 @@ final class CacheAdvanceTests: XCTestCase {
         // Make sure the header data is persisted before we read it as part of the `messages()` call below.
         try header.synchronizeHeaderData()
         // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
-        try header.updateOffsetInFileAtEndOfNewestMessage(to: header.offsetInFileAtEndOfNewestMessage + 1)
+        let offsetInFileAtEndOfNewestMessage = header.offsetInFileAtEndOfNewestMessage
+        + offsetInFileAtEndOfNewestMessageToAdd
+        - offsetInFileAtEndOfNewestMessageToMinus
+        try header.updateOffsetInFileAtEndOfNewestMessage(to: offsetInFileAtEndOfNewestMessage)
         return header
     }
 
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -125,6 +125,80 @@ final class CacheAdvanceTests: XCTestCase {
         }
     }
 
+    private func prepareCacheWithIncorrectOffsetInHeader() throws -> CacheHeaderHandle {
+        let randomHighValue: UInt64 = 10_1000
+        let header = try CacheHeaderHandle(
+            forReadingFrom: testFileLocation,
+            maximumBytes: randomHighValue,
+            overwritesOldMessages: true)
+        let cacheToWrite = CacheAdvance<TestableMessage>(
+            fileURL: testFileLocation,
+            writer: try FileHandle(forWritingTo: testFileLocation),
+            reader: try CacheReader(forReadingFrom: testFileLocation),
+            header: try CacheHeaderHandle(
+                forReadingFrom: testFileLocation,
+                maximumBytes: header.maximumBytes,
+                overwritesOldMessages: header.overwritesOldMessages),
+            decoder: JSONDecoder(),
+            encoder: JSONEncoder())
+
+        let message = TestableMessage(stringLiteral: TestableMessage.lorumIpsum[0].value)
+        try cacheToWrite.append(message: message)
+
+        // Make sure the header data is persisted before we read it as part of the `messages()` call below.
+        try header.synchronizeHeaderData()
+        // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
+        // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
+        try header.updateOffsetInFileAtEndOfNewestMessage(to: header.offsetInFileAtEndOfNewestMessage + 1000)
+        return header
+    }
+
+    func test_messages_timeout_WhenOffsetInFileAtEndOfNewestMessageLargerThanActualFileLength() throws {
+        do {
+            let header = try prepareCacheWithIncorrectOffsetInHeader()
+            let cacheToRead = try CacheAdvance<TestableMessage>(fileURL: testFileLocation,
+                                                                maximumBytes: header.maximumBytes,
+                                                                shouldOverwriteOldMessages: header.overwritesOldMessages)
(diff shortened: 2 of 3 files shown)


Sources/CacheAdvance/CacheReader.swift
```diff
@@ -64,6 +65,11 @@ final class CacheReader {
                 // If the previous read was also empty, then the file has been corrupted.
                 throw CacheAdvanceError.fileCorrupted
             }
+            if seekForwardOnly && startingOffset > FileHeader.expectedEndOfHeaderInFile {
+                // If seekForwardOnly is true, then return nil when startingOffset is ahead
+                // Otherwise, there will be an infinite loop in reading messages.
```
dfed, line 70: I'm not understanding how `seekForwardOnly` doesn't lead to data loss when reading a file where `shouldOverwriteOldMessages` is `true`.

If `shouldOverwriteOldMessages` is `true`, then we might have started reading messages at an offset in the file after `FileHeader.expectedEndOfHeaderInFile`. If `seekForwardOnly` is `true`, then we'll never read the messages between `FileHeader.expectedEndOfHeaderInFile` and `offsetInFileAtEndOfNewestMessage`, right?

Data loss is problematic. I think it'd be preferable to detect that we're in an infinite loop and abort rather than asking customers of this tool to expect data loss in certain circumstances.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -64,6 +65,11 @@ final class CacheReader {
...
+                return nil
```
dfed, line 71: since we're returning early here, can we use a `guard` rather than an `if`? That way it's more clear to future engineers that this block of code will always return

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -125,6 +125,80 @@ final class CacheAdvanceTests: XCTestCase {
...
+                } catch {
+                    XCTFail()
+                }
+            }
+        } catch {
+            XCTFail()
+        }
+    }
+
+
```
dfed, line 201: nit, but let's remove this excess newline:
```suggestion
```

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -125,6 +125,80 @@ final class CacheAdvanceTests: XCTestCase {
         }
     }
 
+    private func prepareCacheWithIncorrectOffsetInHeader() throws -> CacheHeaderHandle {
```
dfed, line 128: Since this is doing pretty complicated work, can you add a `///` comment to explain what this method does?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -125,6 +125,80 @@ final class CacheAdvanceTests: XCTestCase {
...
+        let randomHighValue: UInt64 = 10_1000
```
dfed, line 129: Does this high value need to be "high"? Or does it just need to be higher than the length of `TestableMessage.lorumIpsum[0]`? We have a calculation for `requiredByteCount` method in this file – could we use that?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -125,6 +125,80 @@ final class CacheAdvanceTests: XCTestCase {
...
+            decoder: JSONDecoder(),
+            encoder: JSONEncoder())
+
+        let message = TestableMessage(stringLiteral: TestableMessage.lorumIpsum[0].value)
+        try cacheToWrite.append(message: message)
+
+        // Make sure the header data is persisted before we read it as part of the `messages()` call below.
+        try header.synchronizeHeaderData()
+        // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
+        // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
```
dfed, line 151: If I'm understanding this properly, the infinite loop case above can _only_ be caused by file corruption due to a prior version of the library?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -125,6 +125,80 @@ final class CacheAdvanceTests: XCTestCase {
...
+        try header.updateOffsetInFileAtEndOfNewestMessage(to: header.offsetInFileAtEndOfNewestMessage + 1000)
+        return header
+    }
+
+    func test_messages_timeout_WhenOffsetInFileAtEndOfNewestMessageLargerThanActualFileLength() throws {
```
dfed, line 156: I'll review these tests in more depth once I better understand the problem.

dfed (changes requested):
Thank you for digging in here! I've tried to find a similar bug in the past but the issue eluded me.

Can you help me better understand:
1. How we got into this state we're recovering from with this fix
2. If there is a way to recover without data loss

I'm guessing that #52, #58, and #59 were the PRs that prevented the corruption issue you mention from occurring today?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand: Yes I'd love confirmation here too.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -125,6 +125,80 @@ final class CacheAdvanceTests: XCTestCase {
...
+            encoder: JSONEncoder())
+
+        let message = TestableMessage(stringLiteral: TestableMessage.lorumIpsum[0].value)
+        try cacheToWrite.append(message: message)
+
+        // Make sure the header data is persisted before we read it as part of the `messages()` call below.
+        try header.synchronizeHeaderData()
+        // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
+        // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
+        try header.updateOffsetInFileAtEndOfNewestMessage(to: header.offsetInFileAtEndOfNewestMessage + 1000)
```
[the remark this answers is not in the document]
bachand: I would love to understand this too.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand, line 129:
```suggestion
        let randomHighValue: UInt64 = 101_000
```
While we are here, let's fix this and above to be like ^ as I was misreading this as about 10k but it's really about 100k.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand, line 152: Digging in a bit myself here, I confirmed the that value of `header.offsetInFileAtEndOfNewestMessage` is 203.

Screen Shot 2022-10-10 at 4 08 06 PM


Originally when we said in the test that the "offset in file at end of newest message is larger than the actual file length" I thought that the offset was larger than the maximum bytes, i.e. larger than 101,000.

Now I see that we are saying that the offset is larger than the size of the file on disk, if I'm understanding correctly.

bachand (commented):
Thank you @jianjunwoo! I took an initial look through the PR. Like @dfed I'd also like to better understand how we think that we could have gotten into the situation that causes the infinite loop (i.e. if it's due to a crash, where during code execution would the application need to crash to reproduce the bad state).

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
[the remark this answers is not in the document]
jianjunwoo: yes, it works with 1 as well

image

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
jianjunwoo: I copied the code from [test](https://github.com/dfed/CacheAdvance/pull/64/files#diff-9de095b2bcc9ceaefec7d7de77b3a8f1d5b05e4ae82ba0da807b213ea076ebadL100) above: "func test_messages_canReadEmptyCacheThatDoesNotOverwriteOldestMessages" 

@dfed

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
jianjunwoo: OK. Let me add the comment. This part is copied from another existing [test](https://github.com/dfed/CacheAdvance/pull/64/files#diff-9de095b2bcc9ceaefec7d7de77b3a8f1d5b05e4ae82ba0da807b213ea076ebadL100"

Sources/CacheAdvance/CacheReader.swift
jianjunwoo: @dfed  I check the this line of [code](https://github.com/dfed/CacheAdvance/blob/main/Sources/CacheAdvance/CacheAdvance.swift#L123), and find the writer begins after `FileHeader.expectedEndOfHeaderInFile` when `shouldOverwriteOldMessages` is true. It is the same logic for `shouldOverwriteOldMessages is false `, the writer begins after `FileHeader.expectedEndOfHeaderInFile` too.

When `shouldOverwriteOldMessages` is true, the file handle should not seek back either. Otherwise, it will run into an infinite loop as well.

jianjunwoo: When set both `shouldOverwriteOldMessages` and `seekForwardOnly` true, it won't cause data loss for normal cache file.
The `seekForwardOnly` flag is used avoid the reader to run into this loop: 

Sequence of normal cache file reading:
1. file header
2. message 1 header
3. message 1 body
4. message 2 header
5. message 2 body
6. …
7. end of file

Sequence of bad cache file ( offsetInFileAtEndOfNewestMessage incorrect ) reading:
1. file header
2. message 1 header
3. message 1 body
6. got empty data, then goto step  1)
7. loop forever

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
jianjunwoo: >I see that we are saying that the offset is larger than the size of the file on disk
Yes, this is root cause.

jianjunwoo:
@dfed 
> 1. How we got into this state we're recovering from with this fix
I haven't got the answer either.

> 2. If there is a way to recover without data loss
When set `seekForwardOnly` true, there is no data loss. The reader cursor reaches the end of file, but the offset in the header has a larger value than the file length. So there is no data to read any more.

zack2012:
I talked with @jianjunwoo offline and find some interesting thing. 

This pr unit test proves that if header and body do not match, the code will enter infinite loop. So we need find why they don't match.

I guess one possible reason is that `FileHandle.write` method is not atomicity which means it may write part of data and lead head and body don't match.

I search the open source Foundation repo, find `FileHandle.write` method is not atomicity. See https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/FileHandle.swift#L502

```swift
    @available(swift 5.0)
    public func write<T: DataProtocol>(contentsOf data: T) throws {
        guard self != FileHandle._nulldeviceFileHandle else { return }
        
        guard _isPlatformHandleValid else { throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.fileWriteUnknown.rawValue) }
        
        for region in data.regions {
            try region.withUnsafeBytes { (bytes) in
                if let baseAddress = bytes.baseAddress, bytes.count > 0 {
                    try _writeBytes(buf: UnsafeRawPointer(baseAddress), length: bytes.count)
                }
            }
        }
    }

    internal func _writeBytes(buf: UnsafeRawPointer, length: Int) throws {
#if os(Windows)
        var bytesRemaining = length
        while bytesRemaining > 0 {
            var bytesWritten: DWORD = 0
            if !WriteFile(self._handle, buf.advanced(by: length - bytesRemaining), DWORD(bytesRemaining), &bytesWritten, nil) {
                throw _NSErrorWithWindowsError(GetLastError(), reading: false)
            }
            if bytesWritten == 0 {
                throw _NSErrorWithWindowsError(GetLastError(), reading: false)
            }
            bytesRemaining -= Int(bytesWritten)
        }
#else
        var bytesRemaining = length
        while bytesRemaining > 0 {
            var bytesWritten = 0
            repeat {
                bytesWritten = _write(_fd, buf.advanced(by: length - bytesRemaining), bytesRemaining)
            } while (bytesWritten < 0 && errno == EINTR)
            if bytesWritten <= 0 {
                throw _NSErrorWithErrno(errno, reading: false, path: nil)
            }
            bytesRemaining -= bytesWritten
        }
#endif
```

jianjunwoo:
@zack2012  thanks for pointing out this. Agreed this, it should be caused by the file writing in the iOS foundation.

Sources/CacheAdvance/CacheReader.swift
dfed: In [the line of code you linked](https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L123), we're seeking back to `FileHeader.expectedEndOfHeaderInFile` to write new messages, but older messages still start later in the file. When we open this file to read it, we'll start reading at [`header.offsetInFileOfOldestMessage`](https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L196). When `seekForwardOnly` is `true`, we'll never read newer messages between `FileHeader.expectedEndOfHeaderInFile` and `header.offsetInFileAtEndOfNewestMessage`.

Note that:
* `header.offsetInFileOfOldestMessage` is always greater than or equal to `FileHeader.expectedEndOfHeaderInFile`
* `header.offsetInFileOfOldestMessage ` may be greater than `header.offsetInFileAtEndOfNewestMessage` if messages have been overwritten

I think we could be more cautious when reading and `shouldOverwriteOldMessages` is **false** (which if I recall correctly is how Airbnb uses CacheAdvance for logging). When a file doesn't overwrite new messages the file handle should _not_ seek back when reading since there should never be newer data at the beginning of that file. If that non-overwriting file had been corrupted, then [this guard](https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheReader.swift#L48) might not catch the expected end-of-file, and then we could infinite loop.

I think we can patch the "cache with `shouldOverwriteOldMessages == false` is corrupted results in infinite loop when reading messages" problem I just described by throwing `CacheAdvanceError.fileCorrupted` when we encounter an `.emptyRead` on a cache that should not overwrite old messages. I believe we can make this change without introducing new API.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: 🤦‍♂️ sorry I was reviewing too quickly last night. Thank you for calling out what you were doing in a comment here.

I remember writing that test now. Let me re-review this part of the code with that context in mind 🙂

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -652,5 +678,38 @@ final class CacheAdvanceTests: XCTestCase {
             attributes: nil)
     }
 
+    /// Creates a cache file with incorrect offset in header.
+    /// - Parameter messageCount: The number of messages to write to the cache file
+    /// Returns the header handle of the cache.
+    private func prepareIncorrectOffsetInHeaderAndWriteMessage(messageCount: Int) throws -> CacheHeaderHandle {
+        let randomHighValue: UInt64 = 100_000
```
dfed, line 685: This high value was truly chosen at random in the old test which contained this code – it did not matter what the value was because _no messages were written to the file in this test_.

Now that you've changed this code to actually write messages, we should make sure that the size of the cache is appropriate for the parameters passed into this method.

I think this `randomHighValue` should be _added_ to byte size that is correct for the number of messages we're going to write. Maybe instead of a `messageCount` parameter, we could pass in the messages we want the file to contain as an array. Then we could use the `requiredByteCount` to find the appropriate size for those messages, and then `maximumBytes` would be set to that required byte count _plus_ `randomHighValue`.

As written, this method could unintentionally `throw` `CacheAdvanceError.messageLargerThanRemainingCacheSize` if `messageCount` is high. I'd like to avoid that.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -652,5 +678,38 @@ final class CacheAdvanceTests: XCTestCase {
...
+            encoder: JSONEncoder())
+
+        for _ in 0..<messageCount {
+            let message = TestableMessage(stringLiteral: TestableMessage.lorumIpsum[0].value)
+            try cacheToWrite.append(message: message)
+        }
+
+        // Make sure the header data is persisted before we read it as part of the `messages()` call below.
+        try header.synchronizeHeaderData()
+        // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
```
dfed, line 708: Our file is **not** empty here, as we've written messages to this file above.

Sources/CacheAdvance/CacheReader.swift
dfed: It may _also_ be possible to end up in an infinite loop when reading a corrupted cache where `shouldOverwriteOldMessages == true`. But I think we can detect that case + throw `CacheAdvanceError.fileCorrupted` without introducing new API – we can likely do this by keeping track of whether the reader has already seeked to the beginning of the file and if so, throwing if we're seeking back to the beginning of the file again.

That said, I'm also not convinced we need to solve both corrupted cache cases in the same PR.

dfed:
> When set seekForwardOnly true, there is no data loss.

This is true when caches **do not overwrite older messages**, but I'm not convinced this is true when `shouldOverwriteOldMessages == true`. I provided some guidance [here](https://github.com/dfed/CacheAdvance/pull/64#discussion_r992253565) @jianjunwoo – let's discuss in that thread.

dfed:
Great digging @zack2012 – knowing that the underlying write is not atomic is absolutely helpful. It's also worth noting that the message data write and the header write are separate, so we have multiple places where a timely crash could lead to a corrupted cache file.

dfed:
Though I just had another thought: I believe always write the header _after_ we write the message. So even if the write isn't atomic, a crash mid-write shouldn't create a file where the header's last message is beyond the last written message.

Sources/CacheAdvance/CacheReader.swift
jianjunwoo: @dfed  the context provided are really useful. Put together with the atomic issue, we should more clear about the logics.

For this part, I think `offsetInFileAtEndOfNewestMessage` should be always greater if the file is not  corrupted.

> `header.offsetInFileOfOldestMessage` may be greater than `header.offsetInFileAtEndOfNewestMessage` if messages have been overwritten

I read the [code](https://github.com/dfed/CacheAdvance/blob/main/Sources/CacheAdvance/CacheAdvance.swift#L115) when `header.overwritesOldMessages` = true.
If `cacheHasSpaceForNewMessageBeforeEndOfFile` false, the writer will truncate the cache file and continue to append message. In this case, `offsetInFileAtEndOfNewestMessage` should be greater. And if I`cacheHasSpaceForNewMessageBeforeEndOfFile` is true, the writer continue to append message, `offsetInFileAtEndOfNewestMessage` should be greater too.

jianjunwoo: I'd like to have a new parameter for safety, and this could enable me to have an experiment to see the treatment group in Airbnb App. 
> I believe we can make this change without introducing new API.

jianjunwoo: > I think we could be more cautious when reading and shouldOverwriteOldMessages is false (which if I recall correctly is how Airbnb uses CacheAdvance for logging).

yes, it is false in Airbnb app.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
jianjunwoo: good idea to pass the messages as parameters, let me do that.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -97,34 +97,60 @@ final class CacheAdvanceTests: XCTestCase {
     }
 
     func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
```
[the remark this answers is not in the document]
dfed: Given the above, I would like to better understand how the file corruption you believe you are experiencing could occur. That said, I do still think it is valuable to introduce checks to avoid the infinite loop problem on big-endian machines.

jianjunwoo:
> Though I just had another thought: I believe always write the header _after_ we write the message. So even if the write isn't atomic, a crash mid-write shouldn't create a file where the header's last message is beyond the last written message.

The `offsetInFileAtEndOfNewestMessage` to write to the file header may be not fully written when got crash in this [line](https://github.com/dfed/CacheAdvance/blob/main/Sources/CacheAdvance/CacheHeaderHandle.swift#L74) 
Maybe, only one byte of `offsetInFileAtEndOfNewestMessage` is written to the file. This is caused as the FileHandle.write(...) func is not atomic.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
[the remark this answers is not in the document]
jianjunwoo: Agree about the this. 

> Since [iOS runs in little-endian mode](https://bignerdranch.com/blog/its-the-endian-of-the-world-as-we-know-it/), and the issue you're encountering is occurring on iOS, I do not think we've as-of-yet identified how file corruption would occur.

This is my first time go deep with the integer representation in a real code. 

In little-endian mode, it should be possible to cause the number to be smaller than the last value when crashed when writing.

The`offsetInFileAtEndOfNewestMessage` to write is 
`00002000`. 
The old value in the header is 
`00101000`.
If the write action crash when only the first bytes in `00002000` are written, 
Then we got 
`00001000` 
which is smaller than `00101000` in little-endian mode. 

From this, I added another test with smaller value in the header, it runs into the infinite loop too.
`  func test_messages_timeout_WhenOffsetInFileAtEndOfNewestMessageSmallThanActualFileLength() throws {
`
@dfed

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -97,34 +97,95 @@ final class CacheAdvanceTests: XCTestCase {
...
+            XCTFail()
+        }
+        let expectation = XCTestExpectation()
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
+            expectation.fulfill()
+        }
+        wait(for: [expectation], timeout: 2)
+    }
+
+  func test_messages_timeout_WhenOffsetInFileAtEndOfNewestMessageSmallThanActualFileLength() throws {
```
jianjunwoo, line 137: @dfed  add another test with small offset value. It will cause the infinite loop too as we check like this.
```
        guard startingOffset != offsetInFileAtEndOfNewestMessage else {
            // We're at the last message.
            return nil
        }

```

Sources/CacheAdvance/CacheReader.swift
jianjunwoo: > If you want to ship a version of this repo with the Airbnb app where you can turn this new functionality off, I recommend shipping a fork of this repo.

Emm, that's a better way to do. Really thanks for your careful review and comments!

jianjunwoo: I see there is a line that resets the `offsetInFileOfOldestMessage` value. And then it is synced to the header.
```
                // We know the oldest message is at the beginning of the file, since we are about to toss out the rest of the file.
                reader.offsetInFileOfOldestMessage = FileHeader.expectedEndOfHeaderInFile
                try reader.seekToBeginningOfOldestMessage()

...
            // Update the offsetInFileOfOldestMessage in our header before we write the message.

```
Do I have the right understanding about this line?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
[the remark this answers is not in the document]
dfed: `00002000` is not valid binary: the bits written to disk are only `1`s and `0`s`.

After reviewing the test, I see that we enter an infinite loop because the `offsetInFileAtEndOfNewestMessage` is corrupted! Good catch. I was wrong when I wrote:

> I think in this case we would likely end up reading a [partial message for the last message](https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheReader.swift#L48), which would result in reading data that would [fail to decode](https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L174), leading to the `messages()` call throwing a decoding error

That statement was wrong because we don't validate that the `messageLength` in `.span(messageLength)` does not exceed `offsetInFileAtEndOfNewestMessage`!! We should fix this @jianjunwoo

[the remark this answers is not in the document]
jianjunwoo: > We should fix this
Agree, we could fix this so that the infinite loop should never happen.
But, I still want a parameter for experiment :D . Let me have a fork repo do this. Do you think it's fine? @bachand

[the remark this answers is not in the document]
dfed: Given all of the above, I think we have identified that we should make the following changes:
1. Add a check above [this line](https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheReader.swift#L55) that checks to ensure our read `messageLength` is valid given what we know about the `offsetInFileAtEndOfNewestMessage` and `maximumBytes`. This `guard` would look something like `(startingOffset < offsetInFileAtEndOfNewestMessage && startingOffset + messageLength <= offsetInFileAtEndOfNewestMessage) || startingOffset >= offsetInFileAtEndOfNewestMessage && startingOffset + messageLength < maximumBytes` (though we should likely write this over multiple lines to make it more clear). If this guard fails, we'll `throw CacheAdvanceError.fileCorrupted`.
2. Add a new `guard` [above this line](https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheReader.swift#L63) that the current cache overwrites new messages. If that `guard` does not pass, we should `throw CacheAdvanceError.fileCorrupted`
3. Add a check to the `CacheReader`'s `nextEncodedMessage(...)` method that ensures we don't seek to the beginning of the file more than once.

These changes can be done in the same PR or separate PRs.

As an aside, I think there's an open question regarding whether we want to attach the messages we were able to read before detecting file corruption to the thrown error. Today we don't, but that could be a nice improvement. This kind of improvement could be made on a separate PR as well.

Sources/CacheAdvance/CacheReader.swift
dfed: You're right that on this line we set the `reader`'s `offsetInFileOfOldestMessage` to `expectedEndOfHeaderInFile`:
https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L126-L127

However, the intermediary lines covered by `...` are key here. On this line (part of your first `...`) we seek the reader forward to beginning of the next message _beyond_ the `expectedEndOfHeaderInFile`.
https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L129-L130

Then (part of your second `...`) we `prepareReaderForWriting`:
https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L133-L134

Preparing the reader for writing seeks the reader further if necessary:
https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L209-L222

Once we've done that, we pull out the current reader's `offsetInFile` and assign that value to a local `offsetInFileOfOldestMessage` property:
https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L136-L137

So in conclusion, what we're doing is reseting the `offsetInFileOfOldestMessage` to `expectedEndOfHeaderInFile`, then seeking `offsetInFileOfOldestMessage` forward significantly, and _then_ syncing the new `offsetInFileOfOldestMessage` to the header.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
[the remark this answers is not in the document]
dfed: If you'd like, I'd be happy to write the PRs that implement the above checks. I'm also happy to review PRs that accomplish the above. You + Zack already done a great job of finding issues in this library: your reward for that can be reviewing a PR I write rather than having to do the work yourself 😉. Let me know what you have time to contribute and where I can help!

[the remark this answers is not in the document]
jianjunwoo: Looking forward to the checks working! Go ahead pls : ) 

I would like to fork one, and have a quick fix for the treatment group.
The `shouldOverwriteOldMessages` is false in Airbnb, so it should be safe to apply my quick fix.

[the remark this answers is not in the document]
dfed: Working on implementing the above. I think I've found some simpler checks than what I outlined above. PR incoming once I iron out a few more tests 😄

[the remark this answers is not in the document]
dfed: I believe we can close this PR in favor of #66, as that PR addresses the issues discovered here.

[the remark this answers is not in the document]
bachand: Regarding

> If we crash during step 1, the the next time we open the file we could start reading at an effectively random point in the file. It is highly likely that any "message" read in this file would [fail to decode](https://github.com/dfed/CacheAdvance/blob/0e63e448850a0d64a2a228d59d293749c3c1a248/Sources/CacheAdvance/CacheAdvance.swift#L174), leading to the `messages()` call throwing a decoding error. A crash during step 1 would therefore not result in an infinite reading loop.

it's possible for someone to write a custom message encoder/decoder. If someone were to write a message encoder/decoder that wrote/read raw data, then the encoding would never fail.

Does this change our opinion about the risk of an infinite loop with (1)?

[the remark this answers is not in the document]
dfed: If the decoder never fails, then we have to consider the following failure cases:
1. The crash occurred while writing the message's length (e.g. the `messageSpan`)
2. The crash occurred while writing the message's content (e.g. the `data`)

If the message length didn't write correctly, then we'll end up reading effectively random values for the message length on future "messages". We'll eventually hit a case I've outlined in #66 where a message's length ends after `offsetInFileAtEndOfNewestMessage` and we throw a corrupted file error.

If the message data didn't write correctly, then we'll send the bad message to consuming application, but I don't see how bad message data would cause us to infinite loop.

Sources/CacheAdvance/CacheReader.swift
bachand: Re: https://github.com/dfed/CacheAdvance/pull/64#discussion_r992451299 the `end of newest` should point to the end of `message7` not the start, right? I want to make sure I'm not missing something.
Screen Shot 2022-10-11 at 6 23 17 PM

dfed: You're right @bachand. I've updated the comment. Appreciate the close read!

dfed:
I really appreciate the help of everyone contributing to this PR and the investigation that led to the different test cases here! This investigation has helped explain how we can make this repository _significantly_ more robust in face of crashes while writing.

That said, given that #66 now exists, do we want to continue here or should we close this PR in favor of #66? Let me know how you'd like to proceed.

dfed:
Closing the loop: #66 is merged and 1.2.4 is now released. Let me know how your experiment goes, and I'd love to hear if the eventual adoption of 1.2.4 helps prevent OOMs!

jianjunwoo:
> Closing the loop: #66 is merged and 1.2.4 is now released. Let me know how your experiment goes, and I'd love to hear if the eventual adoption of 1.2.4 helps prevent OOMs!

@dfed  the fix works! it reduces total OOM errors by approximately 70% compared to control group.

dfed:
> > Closing the loop: #66 is merged and 1.2.4 is now released. Let me know how your experiment goes, and I'd love to hear if the eventual adoption of 1.2.4 helps prevent OOMs!
> 
> 
> 
> @dfed  the fix works! it reduces total OOM errors by approximately 70% compared to control group.

Amazing!! Thank you for the update @jianjunwoo. Appreciate your persistence here over the last year+. This wouldn't have gotten fixed without your work!

PR #66, opened 2022-10-11
Protect against crashes during header write

dfed:
If we crash while writing a header, we can end up in a state where our offsetInFileAtEndOfNewestMessage is invalid. Our reading code did not protect against an invalid offsetInFileAtEndOfNewestMessage prior to this change, which could result in an infinite loop while reading messages.

This PR was inspired by https://github.com/dfed/CacheAdvance/pull/64 and https://github.com/dfed/CacheAdvance/pull/64#discussion_r992395768.


Sources/CacheAdvance/CacheReader.swift
```diff
@@ -42,7 +45,7 @@ final class CacheReader {
     }
 
     /// Returns the next encodable message, seeking to the beginning of the next message.
-    func nextEncodedMessage(previousReadWasEmpty: Bool = false) throws -> Data? {
```
dfed, line 45: The case we previously caught with `previousReadWasEmpty` is now caught by our new checks below!

Sources/CacheAdvance/FileHeader.swift
```diff
@@ -79,7 +79,7 @@ struct FileHeader {
     static let version: UInt8 = 1
 
     /// Calculates the offset in the file where the header should end.
-    static var expectedEndOfHeaderInFile = Field(rawValue: Field.allCases.endIndex)!.expectedEndOfFieldInFile
+    static let expectedEndOfHeaderInFile = Field(rawValue: Field.allCases.endIndex)!.expectedEndOfFieldInFile
```
dfed, line 82: This change is unrelated to the overall PR but I saw it and couldn't unsee it.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,20 +96,19 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
-    func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
-        let randomHighValue: UInt64 = 10_1000
```
dfed, line 100: oops. Thanks Michael for catching this one in #64!

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,20 +96,19 @@ final class CacheAdvanceTests: XCTestCase {
...
             fileURL: testFileLocation,
             writer: try FileHandle(forWritingTo: testFileLocation),
-            reader: try CacheReader(forReadingFrom: testFileLocation),
-            header: try CacheHeaderHandle(
+            reader: try CacheReader(
                 forReadingFrom: testFileLocation,
-                maximumBytes: header.maximumBytes,
-                overwritesOldMessages: header.overwritesOldMessages),
+                maximumBytes: randomHighValue),
+            header: header,
```
dfed, line 111: oops. we were previously re-creating the header rather than using the one we already had.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -23,9 +23,12 @@ final class CacheReader {
 
     /// Creates a new instance of the receiver.
     ///
-    /// - Parameter file: The file URL indicating the desired location of the on-disk store. This file should already exist.
-    init(forReadingFrom file: URL) throws {
+    /// - Parameters:
+    ///   - file: The file URL indicating the desired location of the on-disk store. This file should already exist.
+    ///   - maximumBytes: The maximum size of the cache, in bytes. Logs larger than this size will fail to append to the store.
```
bachand, line 28: Let's update this comment. We're in a reader so we're not appending in this class.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -52,6 +55,18 @@ final class CacheReader {
 
         switch try nextEncodedMessageSpan() {
         case let .span(messageLength):
+            // Check our assumptions before we try to read the message.
+            let endOfMessage = Bytes(startingOffset) + Bytes(MessageSpan.storageLength) + Bytes(messageLength)
```
bachand, line 59: Can you help me understand why we need/want to cast to `Bytes`?

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -52,6 +55,18 @@ final class CacheReader {
...
+            let startingOffsetIsBeforeEndOfNewestMessageAndDoesNotExceedEndOfNewestMessage = startingOffset < offsetInFileAtEndOfNewestMessage && endOfMessage <= offsetInFileAtEndOfNewestMessage
+            let startingOffsetIsOnOrAfterAfterEndOfNewestMessageAndDoesNotExceedEndOfFile = startingOffset >= offsetInFileAtEndOfNewestMessage && endOfMessage <= maximumBytes
```
bachand, line 61: We've already guarded that `startingOffset` is not equal to the offset at the end of the newest message. https://github.com/dfed/CacheAdvance/blob/0b63b1e881f7ce12ff4a7dbe9661a696201e0da9/Sources/CacheAdvance/CacheReader.swift#L51

I wonder if that's an opportunity to simplify here since the starting offset should never be on the ned of the newest message.

bachand, line 61:
```suggestion
            let startingOffsetIsOnOrAfterEndOfNewestMessageAndDoesNotExceedEndOfFile = startingOffset >= offsetInFileAtEndOfNewestMessage && endOfMessage <= maximumBytes
```

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,31 +96,108 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
-    func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
-        let randomHighValue: UInt64 = 10_1000
+    func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewestMessageIsBeyondEndOfNewestMessageButBeforeEndOfFile() throws {
+        let message: TestableMessage = "This is a test"
+        let requiredByteCount = try self.requiredByteCount(for: [message])
+        let maximumBytes = requiredByteCount + 2
```
bachand, line 102: Could this be `1`? Trying to make sure I follow.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,31 +96,108 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
-    func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
-        let randomHighValue: UInt64 = 10_1000
+    func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewestMessageIsBeyondEndOfNewestMessageButBeforeEndOfFile() throws {
```
bachand, line 99: Nit
```suggestion
    func test_messages_whenOffsetInFileAtEndOfNewestMessageIsBeyondEndOfNewestMessageButBeforeEndOfFile_throwsFileCorrupted() throws {
```

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,31 +96,108 @@ final class CacheAdvanceTests: XCTestCase {
...
+                    maximumBytes: maximumBytes),
+                header: header,
+                decoder: JSONDecoder(),
+                encoder: JSONEncoder())
+        }
+        let writingCache = try makeCache()
+        try writingCache.append(message: message)
+
+        // Make the file corrupted by setting the offset at end of newest message to be further in the file.
+        // This could happen if a crash occurred during a write of `header.offsetInFileAtEndOfNewestMessage` on a big-endian device.
```
bachand, line 123: I think including the example you already wrote in https://github.com/dfed/CacheAdvance/pull/64#discussion_r992395768 would be very helpful.
Screen Shot 2022-10-11 at 6 18 04 PM

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,31 +96,108 @@ final class CacheAdvanceTests: XCTestCase {
...
+        try header.updateOffsetInFileAtEndOfNewestMessage(
+            to: requiredByteCount + 1)
+
+        // Create a new cache instance that uses the corrupted data persisted to disk
+        let corruptedReadingCache = try makeCache()
+
+        XCTAssertThrowsError(try corruptedReadingCache.messages()) {
```
bachand, line 130: I want to make sure I follow: what was the behavior before this test?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
bachand: The answer is no since we want the offset to be before the end of the file.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,31 +96,108 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
-    func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
```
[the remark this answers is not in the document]
bachand: What I'm understanding is this deleted test is basically a special case of `test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewestMessageIsBeyondEndOfNewestMessageButBeforeEndOfFile` where we had no messages. Is that correct?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,31 +96,108 @@ final class CacheAdvanceTests: XCTestCase {
...
+        // This could happen if a crash occurred during a write of `header.offsetInFileAtEndOfNewestMessage` on a little-endian device.
         try header.updateOffsetInFileAtEndOfNewestMessage(
-            to: FileHeader.expectedEndOfHeaderInFile + 1)
+            to: header.offsetInFileAtEndOfNewestMessage - 1)
+
+        // Create a new cache instance that uses the corrupted data persisted to disk
+        let corruptedReadingCache = try makeCache()
 
-        XCTAssertThrowsError(try cache.messages()) {
+        XCTAssertThrowsError(try corruptedReadingCache.messages()) {
```
bachand, line 200: To make sure I follow again... what was the behavior here before we updated the library?

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -60,15 +75,16 @@ final class CacheReader {
             return message
 
         case .emptyRead:
-            guard !previousReadWasEmpty else {
-                // If the previous read was also empty, then the file has been corrupted.
+            guard !(startingOffset < offsetInFileAtEndOfNewestMessage) else {
```
bachand, line 78: This is cleaner than `previousReadWasEmpty`.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -52,6 +55,18 @@ final class CacheReader {
...
+            // Check our assumptions before we try to read the message.
+            let endOfMessage = Bytes(startingOffset) + Bytes(MessageSpan.storageLength) + Bytes(messageLength)
+            let startingOffsetIsBeforeEndOfNewestMessageAndDoesNotExceedEndOfNewestMessage = startingOffset < offsetInFileAtEndOfNewestMessage && endOfMessage <= offsetInFileAtEndOfNewestMessage
+            let startingOffsetIsOnOrAfterAfterEndOfNewestMessageAndDoesNotExceedEndOfFile = startingOffset >= offsetInFileAtEndOfNewestMessage && endOfMessage <= maximumBytes
+            guard
+                startingOffsetIsBeforeEndOfNewestMessageAndDoesNotExceedEndOfNewestMessage
+                    || startingOffsetIsOnOrAfterAfterEndOfNewestMessageAndDoesNotExceedEndOfFile
+            else {
+                // The offsetInFileAtEndOfNewestMessage is incorrect. This likely occured due to a crash when writing our header file.
+                throw CacheAdvanceError.fileCorrupted
```
bachand, line 67: What do you think about throwing two errors to ensure that we have a test to cover each boolean separately? As-written we can't be as confident that our tests are covering all edge cases.

bachand (commented):
I took a first pass over this @dfed . It takes me awhile to load all of this back in my brain 😅 . Looking real good. I plan to return tomorrow AM if it's ready for another review then.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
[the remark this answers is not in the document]
dfed: Correct!

Sources/CacheAdvance/CacheReader.swift
dfed: `startingOffset` is `UInt64`
 `MessageSpan.storageLength` is `Int`
`messageLength` is `MessageSpan` (i.e. `UInt32`)

We need a common type, and `Bytes` (i.e. `UInt64`) seemed like the right move. That said, I can see an argument for utilizing `UInt64` directly. Let me do that.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: Open to ways to make this more obvious. I was hoping the `+ 2` and `+ 1` in the same test would be reasonably intuitive... but it does require the reader to read the entire method before understanding the magic numbers as-is

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: I should have left a comment! The behavior of this test before the fix was an infinite loop – this test would not complete.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
dfed: I should have left a comment! The behavior of this test before the fix was an infinite loop – this test would not complete.

Sources/CacheAdvance/CacheReader.swift
dfed: 🤦‍♂️

dfed: I like that. Would help us get rid of the `=` check

Sources/CacheAdvance/CacheReader.swift
dfed: I played with this in a debugger – we're not actually catching any issues with `startingOffsetIsAfterEndOfNewestMessageAndDoesNotExceedEndOfFile` (slightly updated name given the commit I just pushed, but you get the idea). What we _are_ doing, however, is ensuring that we pass the `guard` when `offsetInFileAtEndOfNewestMessage < startingOffset`.

Why aren't we catching issues with `startingOffsetIsAfterEndOfNewestMessageAndDoesNotExceedEndOfFile`? Because it's not a unicorn case to hit due to a crash. The only way for the `endOfMessage <= maximumBytes` check in `startingOffsetIsAfterEndOfNewestMessageAndDoesNotExceedEndOfFile` to be `false` is for `header.maximumBytes` to be smaller than expected, and we check that the `header.maximumBytes` in `canWriteToFile`:
https://github.com/dfed/CacheAdvance/blob/2bfee6c0f668735368bf8337cd726928002b1194/Sources/CacheAdvance/CacheHeaderHandle.swift#L160-L174

Granted, we only check `canWriteToFile` when we're appending messages, but since `header.maximumBytes` is never written outside of the initial header write... this seems not worth writing a test against? If our initial header write fails we have other problems.

Sources/CacheAdvance/CacheReader.swift
jianjunwoo, line 78:
```suggestion
            guard startingOffset >= offsetInFileAtEndOfNewestMessage else {
```

jianjunwoo: In this condition, so we will seek to `FileHeader.expectedEndOfHeaderInFile` when `startingOffset >= offsetInFileAtEndOfNewestMessage`. Why do we need to go back? Is it related with `overwritesOldMessages == true`?

dfed: I went back and forth on whether to use `!` here – I appreciate your weighing in. I'll change it 🙂

We check `startingOffset >= offsetInFileAtEndOfNewestMessage` here because if `startingOffset < offsetInFileAtEndOfNewestMessage`, we expect to be reading a message! If we read empty data under this condition, we know that our `offsetInFileAtEndOfNewestMessage` is incorrect, or that a message we read earlier wrote an incorrect message length – i.e. we know that the file is corrupted.

This check is implicitly related to `overwritesOldMessages == true`. As we discussed in #64, when `overwritesOldMessages == false`, we should never seek back to the beginning of the file. On this PR, when `overwritesOldMessages == false` we will not seek back due to a corrupted file: either this check will catch a corrupted file, or this check will:
https://github.com/dfed/CacheAdvance/blob/4d6b753ae2b9bbd837dd7a03ddab50b4b4b0eac0/Sources/CacheAdvance/CacheReader.swift

You can prove the above to yourself by setting `overwritesOldMessages: false` in the unit tests I wrote on this PR.

jianjunwoo: Got it. Now I got fully understanding for `overwritesOldMessages == true` : )

Sources/CacheAdvance/CacheReader.swift
bachand: If I'm understanding correctly the only way `startingOffsetIsAfterEndOfNewestMessageAndDoesNotExceedEndOfFile` could catch issues is in the event of bit flip from a cosmic ray, or similar on-disk data corruption. Does that match your understanding?

bachand (approved):
Thanks @dfed ! Great work.

Sources/CacheAdvance/CacheReader.swift
dfed: Oops! Updated above, and linking here:
https://github.com/dfed/CacheAdvance/blob/fd03b6d0d6584e8a66c90ca77acbbf4db23d50bc/Sources/CacheAdvance/CacheReader.swift#L58-L68

PR #67, opened 2022-10-11
Get codecov working again

dfed:
PRs are no longer getting code coverage reports. It [looks like](https://app.codecov.io/gh/dfed/CacheAdvance/commits) we stopped successfully uploading code coverage reports when we moved to Github Actions:
image

I'm not sure how that PR broke things, but running locally I see that test runs utilizing swift packages have a `.framework` suffix, and test runs utilizing a generated xcodeproj do not. Fixing this seems to have fixed codecov.

PR #68, opened 2022-10-11
Utilize codecov/codecov-action@v3

dfed:
This PR gets us utilizing the newest version of codecov.

Coverage had declined 2+% from `main` since we are now ignoring paths we previously did not ignore. However, ignoring these files is appropriate: we do need to be including `Tests/` or test-only APIs in our calculations. In order to combat the decline in coverage, I removed some unreachable code + did some finagling to add tests to previously untested code.

Relevant resources:
* [Github actions environment variable documentation](https://docs.github.com/en/actions/learn-github-actions/environment-variables)
* [Codecov argument documentation](https://github.com/codecov/codecov-action#arguments)
* [Ignoring paths documentation](https://docs.codecov.com/docs/ignoring-paths)
* [Notice that the `bash` uploader we were using is deprecated](https://docs.codecov.com/docs/supported-languages)
* [PR that added codecov/codecove-action support for Xcode](https://github.com/codecov/codecov-action/pull/699)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: macOS-11
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
@@ -24,7 +24,7 @@ jobs:
     runs-on: macOS-11
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
@@ -45,7 +45,7 @@ jobs:
       fail-fast: false
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
@@ -54,9 +54,6 @@ jobs:
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
         run: Scripts/build.swift ${{ matrix.platforms }}
-      - name: Upload Coverage Reports
-        if: success()
-        run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
   spm-12:
     name: Build Xcode 12
     runs-on: macOS-11
@@ -69,7 +66,7 @@ jobs:
       fail-fast: false
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
@@ -78,9 +75,6 @@ jobs:
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
         run: Scripts/build.swift ${{ matrix.platforms }}
-      - name: Upload Coverage Reports
-        if: success()
-        run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
   spm-13:
     name: Build Xcode 13
     runs-on: macOS-12
@@ -93,7 +87,7 @@ jobs:
       fail-fast: false
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
@@ -102,15 +96,17 @@ jobs:
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
         run: Scripts/build.swift ${{ matrix.platforms }}
+      - name: Prepare Coverage Reports
+        run: ./Scripts/prepare-coverage-reports.sh
       - name: Upload Coverage Reports
         if: success()
-        run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
+        uses: codecov/codecov-action@v3
   spm-13-swift:
     name: Swift Build Xcode 13
     runs-on: macOS-12
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
--- a/Scripts/prepare-coverage-reports.sh
+++ b/Scripts/prepare-coverage-reports.sh
@@ -1,0 +1,35 @@
+#!/bin/zsh -l
+set -e
+
+function exportlcov() {
+    build_type=$1
+    executable_name=$2
+
+    executable=$(find "${directory}" -type f -name $executable_name)
+    profile=$(find "${directory}" -type f -name 'Coverage.profdata')
+    output_file_name="$executable_name.lcov"
+
+    can_proceed=true
+    if [[ $build_type == watchOS* ]]; then
+        echo "\tAborting creation of $output_file_name – watchOS not supported."
+    elif [[ -z $profile ]]; then
+        echo "\tAborting creation of $output_file_name – no profile found."
+    elif [[ -z $executable ]]; then
+        echo "\tAborting creation of $output_file_name – no executable found."
+    else
+        output_dir=".build/artifacts/$build_type"
+        mkdir -p $output_dir
+
+        output_file="$output_dir/$output_file_name"
+        echo "\tExporting $output_file"
+        xcrun llvm-cov export -format="lcov" $executable -instr-profile $profile > $output_file
+    fi
+}
+
+for directory in $(git rev-parse --show-toplevel)/.build/derivedData/*/; do
+    build_type=$(basename $directory)
+    echo "Finding coverage information for $build_type"
+
+    exportlcov $build_type 'CacheAdvanceTests'
+    exportlcov $build_type 'CADCacheAdvanceTests'
+done
--- a/Sources/CADCacheAdvance/CADCacheAdvance.swift
+++ b/Sources/CADCacheAdvance/CADCacheAdvance.swift
@@ -98,30 +98,17 @@ public final class __ObjectiveCCompatibleCacheAdvanceWithGenericData: NSObject {
 /// A decoder that treats all messages as if they are `Data`.
 final class PassthroughDataDecoder: MessageDecoder {
     func decode<T>(_ type: T.Type, from data: Data) throws -> T where T : Decodable {
-        if let data = data as? T {
-            return data
-        } else {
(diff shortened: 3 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: macOS-11
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
dfed, line 15: This change is orthogonal but I noticed warnings in [this build](https://github.com/dfed/CacheAdvance/actions/runs/3230063689) (search for `warnings` on that page) and decide to fix them while I was here.

.github/workflows/ci.yml
```diff
@@ -54,9 +54,6 @@ jobs:
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
         run: Scripts/build.swift ${{ matrix.platforms }}
-      - name: Upload Coverage Reports
-        if: success()
-        run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
```
dfed, line 59: I was having trouble getting Xcode 11 exporting `lcov` file reports, so I deleted the uploads rather than investing more time here. We don't need coverage on all Xcode runs.

.github/workflows/ci.yml
```diff
@@ -78,9 +75,11 @@ jobs:
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
         run: Scripts/build.swift ${{ matrix.platforms }}
+      - name: Prepare Coverage Reports
+        run: ./Scripts/prepare-coverage-reports.sh
```
dfed, line 79: Unfortunately, `codecov/codecov-action@v3`'s `xcode_archive_path` parameter does not seem to work in CI (I could get it working locally but not in CI), so instead of pointing CI at our `.xcresult` files, we instead export our own `lcov` files using a script.

This isn't how I'd like to do this, but it works. And I spent a few too many hours attempting other solutions before getting this solution working. At this point: if the reviewer would like us to generate reports a different way... pull requests welcome 😬

.github/workflows/ci.yml
```diff
@@ -78,9 +75,6 @@ jobs:
         run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
       - name: Build and Test Framework
         run: Scripts/build.swift ${{ matrix.platforms }}
-      - name: Upload Coverage Reports
-        if: success()
-        run: Scripts/upload-coverage-reports.sh ${{ matrix.platforms }}
```
dfed, line 83: Similarly, I was having trouble getting Xcode 12 exporting `lcov` file reports, so I deleted the uploads rather than investing more time here. We don't need coverage on all Xcode runs.

codecov.yml
```diff
@@ -5,3 +5,14 @@ comment:
   layout: "reach,diff,flags,tree"
   behavior: default
   require_changes: no
+
+coverage:
+  status:
+    project:
+      default:
+        threshold: 0.25%
```
dfed, line 13: This is the magic line that lets us merge this PR despite our coverage drop. It's the same threshold we've used in other projects, and it seems reasonable here as well.

Sources/CacheAdvance/EncodableMessage.swift
```diff
@@ -40,7 +40,7 @@ struct EncodableMessage<T: Codable> {
     /// The encoded message, prefixed with the size of the message blob.
     func encodedData() throws -> Data {
         let messageData = try encoder.encode(message)
-        guard messageData.count < MessageSpan.max else {
+        guard messageData.count < Size.max else {
```
dfed, line 43: Utilizing a generic here let me run a test where I passed in a message that would throw the error below without that message requiring multiple gigabytes of RAM.

fdiaz (approved):

.github/workflows/ci.yml
bachand: Asking to learn: if we upload multiple coverage reports for the same code like we were doing before, what does that mean in codecov? Does codecov take the union?

codecov.yml
```diff
@@ -5,3 +5,14 @@ comment:
...
+    patch: off
```
bachand, line 14: What does this do? I couldn't find documentation.

Sources/SwiftTryCatch/SwiftTryCatch.m
```diff
@@ -24,19 +24,14 @@ @implementation SwiftTryCatch
 /**
  Provides try catch functionality for swift by wrapping around Objective-C
  */
-+ (void)try:(__attribute__((noescape)) void(^ _Nonnull)(void))try catch:(__attribute__((noescape)) void(^ _Nonnull)(NSException *exception))catch finally:(__attribute__((noescape)) void(^ _Nullable)(void))finally;
```
bachand, line 27: I assume we never used this functionality?

Sources/CADCacheAdvance/CADCacheAdvance.swift
```diff
@@ -98,30 +98,17 @@ public final class __ObjectiveCCompatibleCacheAdvanceWithGenericData: NSObject {
...
     func decode<T>(_ type: T.Type, from data: Data) throws -> T where T : Decodable {
-        if let data = data as? T {
-            return data
-        } else {
-            throw DecodingError.dataCorrupted(
-                DecodingError.Context(
-                    codingPath: [],
-                    debugDescription: "Type was not Data"))
-        }
+        // Force cast because this type is only used with a CacheAdvance<Data> type.
```
bachand, line 101: So it would be programmer error if someone used `PassthroughDataDecoder` with a result type other than `Data`?

Scripts/prepare-coverage-reports.sh
```diff
@@ -0,0 +1,35 @@
...
+    profile=$(find "${directory}" -type f -name 'Coverage.profdata')
+    output_file_name="$executable_name.lcov"
+
+    can_proceed=true
+    if [[ $build_type == watchOS* ]]; then
+        echo "\tAborting creation of $output_file_name – watchOS not supported."
+    elif [[ -z $profile ]]; then
+        echo "\tAborting creation of $output_file_name – no profile found."
+    elif [[ -z $executable ]]; then
+        echo "\tAborting creation of $output_file_name – no executable found."
```
bachand, line 18: Should we fail in these cases?

Scripts/prepare-coverage-reports.sh
```diff
@@ -0,0 +1,35 @@
...
+    else
+        output_dir=".build/artifacts/$build_type"
+        mkdir -p $output_dir
+
+        output_file="$output_dir/$output_file_name"
+        echo "\tExporting $output_file"
+        xcrun llvm-cov export -format="lcov" $executable -instr-profile $profile > $output_file
```
bachand, line 25: Trying to make sure I follow: how did we get this report before and why do we need to get it differently now?

bachand (commented):
Thanks @dfed and @fdiaz for the review. I looked this over and added some questions. I also wanted to know: do we have confirmation that this PR resolved https://github.com/dfed/CacheAdvance/pull/67#issuecomment-1275266242?

codecov.yml
dfed: `patch`? I was copied from SwiftInspector, which introduced this in https://github.com/fdiaz/SwiftInspector/pull/132

Sources/SwiftTryCatch/SwiftTryCatch.m
dfed: Exactly. Wasn't used and was lowering our coverage

Sources/CADCacheAdvance/CADCacheAdvance.swift
dfed: It would. That type is internal to the repo so there's not much chance of that happening.

Scripts/prepare-coverage-reports.sh
dfed: Not necessarily. Since I put `watchOS` at the top we should never hit these cases. But it seems reasonable that we'd run jobs that wouldn't produce code coverage data – we're not being smart about what build types we apply this prepare step on.

.github/workflows/ci.yml
dfed: Yeah codecov unions (they call it "merging") them: https://docs.codecov.com/docs/merging-reports

Scripts/prepare-coverage-reports.sh
dfed: Before we were using codecov's bash uploader to create this script for us. That script is deprecated. The new system fails to find the `xcresult` files that the bash uploader could find (https://github.com/dfed/CacheAdvance/pull/68#discussion_r992933625), so now we have to create a `lcov` file (which the new system understands and can find) ourselves.

Basically, the new system's Xcode support is lacking, so I created a file type it understood natively.

bachand: Ah I missed that comment since it was on an outdated line. Thanks for the pointer 👍

dfed: All good – I should have re-applied that comment!

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: macOS-11
```
bachand, line 12: Putting this here so I have a thread. In my PR review I asked:

> I also wanted to know: do we have confirmation that this PR resolved https://github.com/dfed/CacheAdvance/pull/67#issuecomment-1275266242?

I am having trouble following what exactly we think fixed the codecov issue ultimately and how we're confident that it's resolved. I'm following up because I want to understand better. Thanks for all of your work here @dfed !

dfed: Oh, what fixed the codecov issue was:
1. Updating to a newer (non-deprecated) version of codecov
2. Manually creating the `lcov` file that codecov could pick up (since the newer version of codecov couldn't find the coverage files in derived data either)

Based on the logs from recent builds – the deprecated version of codecov wasn't picking up the `xcresult` file that includes coverage information. I probably could have fixed the issue by manually creating the `lcov` file like I did in this MR, but I figured I should start by updating to a non-deprecated tool.

You can see from [recent logs](https://github.com/dfed/CacheAdvance/actions/runs/3237806944/jobs/5305302300#step:8:30) that we are finding+uploading the `lcov` files. Looking through other recent builds ([all builds are here](https://github.com/dfed/CacheAdvance/actions)), you can see the same logs.

The very high-level is: codecov was failing to find our coverage data with the old config, and moving to this new config (namely, making the `lcov` files for codecov rather than making codecov create these files from our derived data) fixed that.

I still do not understand why neither the deprecated tool nor the modern tool could create these coverage files given the inputs we were giving them. But at least we have a workaround.

bachand: Thank you for explaining, and for the log pointer!

PR #70, opened 2022-10-12
Get CI running on Xcode 14

dfed:
As title. Adds yet another Xcode version to support!

In this PR I also balance out the workloads between runners to lower CI's wall time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,3 +113,41 @@ jobs:
         run: sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
       - name: Build and Test Framework
         run: xcrun swift test -c release -Xswiftc -enable-testing
+  spm-14:
+    name: Build Xcode 14
+    runs-on: macOS-12
+    strategy:
+      matrix:
+        platforms: [
+          'iOS_16,tvOS_16,watchOS_9',
+          'macOS_12',
+        ]
+      fail-fast: false
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: Bundle Install
+        run: bundle install
+      - name: Select Xcode Version
+        run: sudo xcode-select --switch /Applications/Xcode_14.0.1.app/Contents/Developer
+      - name: Prepare Simulator Runtimes
+        run: Scripts/github/prepare-simulators.sh ${{ matrix.platforms }}
+      - name: Build and Test Framework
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -113,3 +113,41 @@ jobs:
...
       - name: Build and Test Framework
         run: xcrun swift test -c release -Xswiftc -enable-testing
+  spm-14:
+    name: Build Xcode 14
+    runs-on: macOS-12
+    strategy:
+      matrix:
+        platforms: [
+          'iOS_16,watchOS_9',
+          'macOS_12,tvOS_16',
```
dfed, line 123: Note that we're testing `macOS_12` here again. We don't _need_ to do this but given that we're running with a different SDK I figure why not.

.github/workflows/ci.yml
```diff
@@ -39,8 +39,8 @@ jobs:
     strategy:
       matrix:
         platforms: [
-          'iOS_13,tvOS_13,watchOS_6',
-          'macOS_10_15',
+          'iOS_13,watchOS_6,macOS_10_15',
+          'tvOS_13',
```
bachand, line 43: Can you help me understand the changes here and below? Why do we need to move things around?

Scripts/build.swift
```diff
@@ -61,19 +68,23 @@ enum Platform: String, CaseIterable, CustomStringConvertible {
             return "OS=7.2,name=Apple Watch Series 6 - 44mm"
         case .watchOS_8:
             return "OS=8.5,name=Apple Watch Series 6 - 44mm"
+        case .watchOS_9:
+            return "OS=9.0,name=Apple Watch Series 7 (45mm)"
```
bachand, line 72: To confirm this friendly name changed from `-` to `(` `)`?

bachand (approved):
Thanks @dfed !

.github/workflows/ci.yml
dfed: We don't _need_ to move these things around, but I'm playing with moving them around to see if I can lower the overall wall time taken to complete a run. Some jobs complete faster than others, and I'm trying to optimize such that the sharded jobs have equal-ish runtimes. I may play with this setup once more before merging – still gathering data.

Scripts/build.swift
dfed: yup. I don't know why but Xcode 14 did change the formatting here.

PR #71, opened 2022-10-12
Set a required ruby version in the Gemfile

dfed:
This PR sets the ruby version in our Gemfile. This should enable engineers working on ARM Macs to utilize `bundle exec pod trunk push` without failing locally. By default, the Gemfile utilizes system Ruby, which unfortunately does not support ARM Macs well.

Cc @paulabrudan


.github/workflows/ci.yml
```diff
@@ -11,6 +11,9 @@ jobs:
     name: Pod Lint
     runs-on: macOS-11
     steps:
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.7.6'
```
dfed, line 16: I'd like to use a newer version of ruby, but this one works and it's the latest supported by the `setup-ruby` script

Gemfile
```diff
@@ -1,3 +1,5 @@
+ruby '2.7.6'
```
bachand, line 1: > By default, the Gemfile utilizes system Ruby, which unfortunately does not support ARM Macs well.

Curious what you mean by this. I'd think/hope the system Ruby shipped on ARM Macs worked OK but maybe that's not the case?

bachand (approved):
Nice!

Gemfile
dfed: What I know: running `bundle exec pod trunk push` fails with Apple's ruby. But with later versions it works.

With Apple's ruby, we fail loading `ffi`, and the stated solution is to bump the ruby version: see https://github.com/ffi/ffi/issues/922

dfed: Maybe macOS 13 resolves this issue? But for now this seems reasonable

bachand: Yep seems very reasonable to enforce a Ruby version 👍

PR #72, opened 2022-11-04
Split two ranges to read messages if offsetInFileOfOldestMessage > offsetInFileAtEndOfNewestMessage to avoid possible loop

jianjunwoo:
My concern is there is still this line of code: [`try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)` ](https://github.com/dfed/CacheAdvance/pull/66/files#diff-90b1f5431bb05caf1445ee7ee42e2a66b2bfb2e6fad63f3266cace6cd1389fd6L68), which keeps the possibility to run into an infinite loop. 

Thus, I have a thought to fix this for `shouldOverwriteOldMessages` true | false without the seek back action in this func. 

1. When `shouldOverwriteOldMessages` is true and new_offset < old_offset , split the reader by two ranges according old message offset, new message offset. 
    1.1 Range 1 would be : from old message offset to file end.
    1.2 Range 2 would be: from top to newest offset
2. When `shouldOverwriteOldMessages` is false,  the reader read from top to end.
3. Both of the above situations, we never seek back for next read which can guarantee there is no loop.

@dfed  @bachand

### DIFF
--- a/CacheAdvance.podspec
+++ b/CacheAdvance.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '1.2.3'
+  s.version  = '1.2.5'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'A performant cache for logging systems. CacheAdvance persists log events 30x faster than SQLite.'
   s.homepage = 'https://github.com/dfed/CacheAdvance'
--- a/CacheAdvance.podspec
+++ b/CacheAdvance.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '1.2.4'
+  s.version  = '1.2.3'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'A performant cache for logging systems. CacheAdvance persists log events 30x faster than SQLite.'
   s.homepage = 'https://github.com/dfed/CacheAdvance'
--- a/Sources/CacheAdvance/CacheAdvance.swift
+++ b/Sources/CacheAdvance/CacheAdvance.swift
@@ -47,9 +47,7 @@ public final class CacheAdvance<T: Codable> {
         self.init(
             fileURL: fileURL,
             writer: try FileHandle(forWritingTo: fileURL),
-            reader: try CacheReader(
-                forReadingFrom: fileURL,
-                maximumBytes: maximumBytes),
+            reader: try CacheReader(forReadingFrom: fileURL),
             header: try CacheHeaderHandle(
                 forReadingFrom: fileURL,
                 maximumBytes: maximumBytes,
--- a/Sources/CacheAdvance/CacheAdvance.swift
+++ b/Sources/CacheAdvance/CacheAdvance.swift
@@ -170,10 +170,28 @@ public final class CacheAdvance<T: Codable> {
         try header.checkFile()
 
         var messages = [T]()
-        while let encodedMessage = try reader.nextEncodedMessage() {
-            messages.append(try decoder.decode(T.self, from: encodedMessage))
-        }
+        // There is only one range: | `offsetInFileOfOldestMessage` -> `offsetInFileAtEndOfNewestMessage`|
+        if reader.offsetInFileOfOldestMessage < reader.offsetInFileAtEndOfNewestMessage {
+            for encodedMessage in try reader.encodedMessagesFromOffset(reader.offsetInFileOfOldestMessage,
+                                                                       endOffset: reader.offsetInFileAtEndOfNewestMessage) {
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+        } else {
+            // In this case, the messages could be split to two ranges
+            // | First Range | (GAP: ignore) | Second Range |
 
+            // This is second range: | `offsetInFileOfOldestMessage` -> EOF |
+            for encodedMessage in try reader.encodedMessagesFromOffset(reader.offsetInFileOfOldestMessage) {
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+
+            // This is first range: | `expectedEndOfHeaderInFile` -> `offsetInFileAtEndOfNewestMessage`|
+            for encodedMessage in try reader.encodedMessagesFromOffset(
+                FileHeader.expectedEndOfHeaderInFile,
+                endOffset: reader.offsetInFileAtEndOfNewestMessage) {
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+        }
         // Now that we've read all messages, seek back to the oldest message.
         try reader.seekToBeginningOfOldestMessage()
 
--- a/Sources/CacheAdvance/CacheReader.swift
+++ b/Sources/CacheAdvance/CacheReader.swift
@@ -41,15 +41,24 @@ final class CacheReader {
         reader.offsetInFile
     }
 
-    /// Returns the next encodable message, seeking to the beginning of the next message.
-    func nextEncodedMessage(previousReadWasEmpty: Bool = false) throws -> Data? {
-        let startingOffset = offsetInFile
-
-        guard startingOffset != offsetInFileAtEndOfNewestMessage else {
-            // We're at the last message.
-            return nil
+    /// Returns the encodable messages in a range
+    func encodedMessagesFromOffset(_ startOffset: UInt64, endOffset: UInt64? = nil) throws -> [Data] {
+        var encodedMessages = [Data]()
+        try reader.seek(to: startOffset)
+        while let data = try nextEncodedMessage() {
+            encodedMessages.append(data)
+            if let endOffset, offsetInFile >= endOffset {
+                break
+            }
         }
+        if let endOffset, offsetInFile != endOffset {
+            throw CacheAdvanceError.fileCorrupted
+        }
+        return encodedMessages
+    }
 
+    /// Returns the next encodable message, seeking to the beginning of the next message.
+    func nextEncodedMessage() throws -> Data? {
         switch try nextEncodedMessageSpan() {
         case let .span(messageLength):
             let message = try reader.readDataUp(toLength: Int(messageLength))
@@ -60,15 +69,7 @@ final class CacheReader {
             return message
 
         case .emptyRead:
-            guard !previousReadWasEmpty else {
-                // If the previous read was also empty, then the file has been corrupted.
-                throw CacheAdvanceError.fileCorrupted
-            }
-            // We know the next message is at the end of the file header. Let's seek to it.
-            try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)
-
-            // We know there's a message to read now that we're at the start of the file.
-            return try nextEncodedMessage(previousReadWasEmpty: true)
+            return nil
 
         case .invalidFormat:
             throw CacheAdvanceError.fileCorrupted
--- a/Sources/CacheAdvance/CacheReader.swift
+++ b/Sources/CacheAdvance/CacheReader.swift
@@ -23,12 +23,9 @@ final class CacheReader {
 
     /// Creates a new instance of the receiver.
     ///
-    /// - Parameters:
-    ///   - file: The file URL indicating the desired location of the on-disk store. This file should already exist.
-    ///   - maximumBytes: The maximum size of the cache, in bytes.
-    init(forReadingFrom file: URL, maximumBytes: Bytes) throws {
+    /// - Parameter file: The file URL indicating the desired location of the on-disk store. This file should already exist.
+    init(forReadingFrom file: URL) throws {
         reader = try FileHandle(forReadingFrom: file)
-        self.maximumBytes = maximumBytes
     }
 
     deinit {
@@ -45,7 +42,7 @@ final class CacheReader {
     }
 
     /// Returns the next encodable message, seeking to the beginning of the next message.
-    func nextEncodedMessage() throws -> Data? {
+    func nextEncodedMessage(previousReadWasEmpty: Bool = false) throws -> Data? {
         let startingOffset = offsetInFile
 
         guard startingOffset != offsetInFileAtEndOfNewestMessage else {
@@ -55,18 +52,6 @@ final class CacheReader {
 
         switch try nextEncodedMessageSpan() {
         case let .span(messageLength):
-            // Check our assumptions before we try to read the message.
-            let endOfMessage = startingOffset + UInt64(MessageSpan.storageLength) + UInt64(messageLength)
-            let startingOffsetIsBeforeEndOfNewestMessageAndDoesNotExceedEndOfNewestMessage = startingOffset < offsetInFileAtEndOfNewestMessage && endOfMessage <= offsetInFileAtEndOfNewestMessage
-            let startingOffsetIsAfterEndOfNewestMessageAndDoesNotExceedEndOfFile = offsetInFileAtEndOfNewestMessage < startingOffset && endOfMessage <= maximumBytes
-            guard
-                startingOffsetIsBeforeEndOfNewestMessageAndDoesNotExceedEndOfNewestMessage
-                    || startingOffsetIsAfterEndOfNewestMessageAndDoesNotExceedEndOfFile
-            else {
-                // The offsetInFileAtEndOfNewestMessage is incorrect. This likely occured due to a crash when writing our header file.
-                throw CacheAdvanceError.fileCorrupted
-            }
-
             let message = try reader.readDataUp(toLength: Int(messageLength))
             guard message.count > 0 else {
                 throw CacheAdvanceError.fileCorrupted
@@ -75,16 +60,15 @@ final class CacheReader {
             return message
 
         case .emptyRead:
-            guard offsetInFileAtEndOfNewestMessage < startingOffset else {
-                // We started reading before the offset of the end of the newest message, therefore we expected a message to be read. We instead read an empty space, meaning that the file is corrupt.
+            guard !previousReadWasEmpty else {
+                // If the previous read was also empty, then the file has been corrupted.
                 throw CacheAdvanceError.fileCorrupted
             }
-
             // We know the next message is at the end of the file header. Let's seek to it.
             try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)
 
             // We know there's a message to read now that we're at the start of the file.
-            return try nextEncodedMessage()
+            return try nextEncodedMessage(previousReadWasEmpty: true)
 
         case .invalidFormat:
             throw CacheAdvanceError.fileCorrupted
@@ -132,7 +116,6 @@ final class CacheReader {
     }
 
     private let reader: FileHandle
-    private let maximumBytes: Bytes
 
 }
 
--- a/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
+++ b/Tests/CacheAdvanceTests/CacheAdvanceTests.swift
@@ -96,114 +96,31 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
-    func test_messages_whenOffsetInFileAtEndOfNewestMessageIsBeyondEndOfNewestMessageButBeforeEndOfFile_throwsFileCorrupted() throws {
-        let message: TestableMessage = "This is a test"
-        let requiredByteCount = try self.requiredByteCount(for: [message])
-        let maximumBytes = requiredByteCount + 2
+    func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
+        let randomHighValue: UInt64 = 10_1000
         let header = try CacheHeaderHandle(
             forReadingFrom: testFileLocation,
-            maximumBytes: maximumBytes,
+            maximumBytes: randomHighValue,
             overwritesOldMessages: true)
+        let cache = CacheAdvance<TestableMessage>(
+            fileURL: testFileLocation,
+            writer: try FileHandle(forWritingTo: testFileLocation),
+            reader: try CacheReader(forReadingFrom: testFileLocation),
+            header: try CacheHeaderHandle(
+                forReadingFrom: testFileLocation,
+                maximumBytes: header.maximumBytes,
+                overwritesOldMessages: header.overwritesOldMessages),
+            decoder: JSONDecoder(),
+            encoder: JSONEncoder())
 
-        func makeCache() throws -> CacheAdvance<TestableMessage> {
-            return CacheAdvance<TestableMessage>(
-                fileURL: testFileLocation,
-                writer: try FileHandle(forWritingTo: testFileLocation),
-                reader: try CacheReader(
-                    forReadingFrom: testFileLocation,
-                    maximumBytes: maximumBytes),
-                header: header,
-                decoder: JSONDecoder(),
-                encoder: JSONEncoder())
-        }
-        let writingCache = try makeCache()
-        try writingCache.append(message: message)
-
-        // Make the file corrupted by setting the offset at end of newest message to be further in the file.
-        // This could happen if a crash occurred during a write of `header.offsetInFileAtEndOfNewestMessage` on a big-endian device.
-        // Big-endian devices write the most significant digits first, meaning that if we were offsetInFileAtEndOfNewestMessage from 00001010 to 00010000, it would be possible to crash with the following bytes written to disk: 00011010.
-        // The 00011010 value is a larger value what we intended to write, which would lead to file corruption.
+        // Make sure the header data is persisted before we read it as part of the `messages()` call below.
+        try header.synchronizeHeaderData()
+        // Our file is empty. Make the file corrupted by setting the offset at end of newest message to be further in the file.
+        // This should never happen, but past versions of this repo could lead to a file having this kind of inconsistency if a crash occurred at the wrong time.
         try header.updateOffsetInFileAtEndOfNewestMessage(
-            to: requiredByteCount + 1)
+            to: FileHeader.expectedEndOfHeaderInFile + 1)
 
-        // Create a new cache instance that uses the corrupted data persisted to disk
-        let corruptedReadingCache = try makeCache()
-
-        XCTAssertThrowsError(try corruptedReadingCache.messages()) {
-            XCTAssertEqual($0 as? CacheAdvanceError, CacheAdvanceError.fileCorrupted)
-        }
-    }
-
-    func test_messages_whenOffsetInFileAtEndOfNewestMessageIsBeyondEndOfFile_throwsFileCorrupted() throws {
-        let message: TestableMessage = "This is a test"
-        let maximumBytes = try requiredByteCount(for: [message])
-        let header = try CacheHeaderHandle(
-            forReadingFrom: testFileLocation,
-            maximumBytes: maximumBytes,
-            overwritesOldMessages: true)
-
-        func makeCache() throws -> CacheAdvance<TestableMessage> {
-            return CacheAdvance<TestableMessage>(
-                fileURL: testFileLocation,
-                writer: try FileHandle(forWritingTo: testFileLocation),
-                reader: try CacheReader(
-                    forReadingFrom: testFileLocation,
-                    maximumBytes: maximumBytes),
-                header: header,
-                decoder: JSONDecoder(),
-                encoder: JSONEncoder())
-        }
-        let writingCache = try makeCache()
-        try writingCache.append(message: message)
-
-        // Make the file corrupted by setting the offset at end of newest message to be further in the file.
-        // This could happen if a crash occurred during a write of `header.offsetInFileAtEndOfNewestMessage` on a big-endian device.
-        // Big-endian devices write the most significant digits first, meaning that if we were offsetInFileAtEndOfNewestMessage from 00001010 to 00010000, it would be possible to crash with the following bytes written to disk: 00011010.
-        // The 00011010 value is a larger value what we intended to write, which would lead to file corruption.
-        try header.updateOffsetInFileAtEndOfNewestMessage(
-            to: header.offsetInFileAtEndOfNewestMessage + 1)
-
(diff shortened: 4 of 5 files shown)


Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -172,10 +170,28 @@ public final class CacheAdvance<T: Codable> {
...
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+
+            // This is first range: | `expectedEndOfHeaderInFile` -> `offsetInFileAtEndOfNewestMessage`|
+            for encodedMessage in try reader.encodedMessagesFromOffset(
+                FileHeader.expectedEndOfHeaderInFile,
+                endOffset: reader.offsetInFileAtEndOfNewestMessage) {
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+        }
```
dfed, line 194: This is a clever approach! I don't have time to dig to deeply into this PR this week, but I'll do my best to get to it next week. My initial impression is that this approach seems reasonable. At a minimum, I really like the code comment you've written here – they'll really help future maintainers (myself included!).

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -172,10 +170,28 @@ public final class CacheAdvance<T: Codable> {
...
+            for encodedMessage in try reader.encodedMessagesFromOffset(reader.offsetInFileOfOldestMessage,
+                                                                       endOffset: reader.offsetInFileAtEndOfNewestMessage) {
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+        } else {
+            // In this case, the messages could be split to two ranges
+            // | First Range | (GAP: ignore) | Second Range |
 
+            // This is second range: | `offsetInFileOfOldestMessage` -> EOF |
+            for encodedMessage in try reader.encodedMessagesFromOffset(reader.offsetInFileOfOldestMessage) {
```
dfed, line 184: We know the `endOffset` here, right? Since we know the `maximumBytes` of the file?

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -44,29 +41,26 @@ final class CacheReader {
...
 
-    /// Returns the next encodable message, seeking to the beginning of the next message.
-    func nextEncodedMessage() throws -> Data? {
-        let startingOffset = offsetInFile
-
-        guard startingOffset != offsetInFileAtEndOfNewestMessage else {
-            // We're at the last message.
-            return nil
+    /// Returns the encodable messages in a range
+    func encodedMessagesFromOffset(_ startOffset: UInt64, endOffset: UInt64? = nil) throws -> [Data] {
```
dfed, line 45: I think we can make `endOffset` required given my previous comment.

I'd also consider renaming `endOffset` here to be `maximumOffset` or something to make it more clear that it is the largest possible offset for the end of this message, but that the message may (or even likely!) end before that offset.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -44,29 +41,26 @@ final class CacheReader {
...
-            let endOfMessage = startingOffset + UInt64(MessageSpan.storageLength) + UInt64(messageLength)
-            let startingOffsetIsBeforeEndOfNewestMessageAndDoesNotExceedEndOfNewestMessage = startingOffset < offsetInFileAtEndOfNewestMessage && endOfMessage <= offsetInFileAtEndOfNewestMessage
-            let startingOffsetIsAfterEndOfNewestMessageAndDoesNotExceedEndOfFile = offsetInFileAtEndOfNewestMessage < startingOffset && endOfMessage <= maximumBytes
-            guard
-                startingOffsetIsBeforeEndOfNewestMessageAndDoesNotExceedEndOfNewestMessage
-                    || startingOffsetIsAfterEndOfNewestMessageAndDoesNotExceedEndOfFile
-            else {
-                // The offsetInFileAtEndOfNewestMessage is incorrect. This likely occured due to a crash when writing our header file.
-                throw CacheAdvanceError.fileCorrupted
-            }
```
dfed, line 68: can you help me understand why we no longer need these checks?

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -75,16 +69,7 @@ final class CacheReader {
...
-                // We started reading before the offset of the end of the newest message, therefore we expected a message to be read. We instead read an empty space, meaning that the file is corrupt.
-                throw CacheAdvanceError.fileCorrupted
-            }
-
-            // We know the next message is at the end of the file header. Let's seek to it.
-            try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)
-
-            // We know there's a message to read now that we're at the start of the file.
-            return try nextEncodedMessage()
+            return nil
```
dfed, line 72: can you add a code comment here to help me understand what case we're hitting here / why it is expected that we return `nil` here?

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,114 +96,31 @@ final class CacheAdvanceTests: XCTestCase {
         XCTAssertEqual(messages, [])
     }
 
-    func test_messages_whenOffsetInFileAtEndOfNewestMessageIsBeyondEndOfNewestMessageButBeforeEndOfFile_throwsFileCorrupted() throws {
```
dfed, line 99: why is this test (and the deleted one below) no longer valid? I think testing these cases is still valuable, but I might be missing something.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
```diff
@@ -96,114 +96,31 @@ final class CacheAdvanceTests: XCTestCase {
...
-        XCTAssertThrowsError(try corruptedReadingCache.messages()) {
-            XCTAssertEqual($0 as? CacheAdvanceError, CacheAdvanceError.fileCorrupted)
-        }
-    }
-
-    func test_messages_whenOffsetInFileAtEndOfNewestMessageIsBeforeEndOfNewestMessage_throwsFileCorrupted() throws {
-        let message: TestableMessage = "This is a test"
-        let maximumBytes = try requiredByteCount(for: [message])
+    func test_messages_throwsFileCorruptedWhenOffsetInFileAtEndOfNewsetMessageOutOfSync() throws {
+        let randomHighValue: UInt64 = 10_1000
```
dfed, line 100: let's make this number less confusing (and fix an old mistake of mine) by doing either:
```suggestion
        let randomHighValue: UInt64 = 10_000
```
or:
```suggestion
        let randomHighValue: UInt64 = 101_000
```

dfed (commented):
I like how you're thinking about this problem! I need to dig in more before I can approve, which I won't have time to do until next week, but I'm excited by this direction 😄. In the meantime, I've left some comments/questions.

CacheAdvance.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
```
dfed, line 1: Starting a thread for this discussion:

> My concern is there is still this line of code: [`try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)` ](https://github.com/dfed/CacheAdvance/pull/66/files#diff-90b1f5431bb05caf1445ee7ee42e2a66b2bfb2e6fad63f3266cace6cd1389fd6L68), which keeps the possibility to run into an infinite loop. 

I don't think that line can lead to an infinite loop per the unit tests in #66. We will either read all messages and then hit the end and stop, or we will detect that the file is corrupted and stop.

That said, I think the approach you're taking here may be easier to maintain + understand than the approach I took in #66, so I'm open to this approach even if we're not solving an existing problem. However, I would feel much more confident with this approach if we didn't revert the tests introduced in #66.

Tests/CacheAdvanceTests/CacheAdvanceTests.swift
jianjunwoo: These tests should be kept. I reverted the last commit, this was deleted then.

CacheAdvance.podspec
dfed: That said, this new code may have different behavior (maybe we don't throw `fileCorrupted` anymore?), so we may need to _update_ the tests from #66.

jianjunwoo: > However, I would feel much more confident with this approach if we didn't revert the tests introduced in https://github.com/dfed/CacheAdvance/pull/66.

That's helpful. Let me keep the tests.

dfed: Per https://github.com/dfed/CacheAdvance/pull/72#discussion_r1013594675, it sounds like we're intending to keep the tests from #66! Excited to see if they need to be modified to pass with this change 🙂

Sources/CacheAdvance/CacheAdvance.swift
jianjunwoo: > I like how you're thinking about this problem! I need to dig in more before I can approve, which I won't have time to do until next week, but I'm excited by this direction 😄. In the meantime, I've left some comments/questions.

This approach came in when I read the discussion in PR#64 again. really happy to see this move forward if you feel it right too : D

Sources/CacheAdvance/CacheAdvance.swift
jianjunwoo: The `maximumBytes` may larger then the end offset I think.

Sources/CacheAdvance/CacheReader.swift
jianjunwoo: I would like to say we are safe if do not seek back in this method.

Sources/CacheAdvance/CacheReader.swift
jianjunwoo: here we return nil to stop reading

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -44,29 +41,29 @@ final class CacheReader {
...
+    func encodedMessagesFromOffset(_ startOffset: UInt64, endOffset: UInt64? = nil) throws -> [Data] {
+        var encodedMessages = [Data]()
+        try reader.seek(to: startOffset)
+        while let data = try nextEncodedMessage() {
+            encodedMessages.append(data)
+            if let endOffset, offsetInFile >= endOffset {
+                break
+            }
+        }
+        if let endOffset, offsetInFile != endOffset {
```
jianjunwoo, line 57: If use `maximumOffset` set default value for `endOffset` in the func's parameter, this line need change to `offsetInFile < endOffset`.  
If changed to this, we need to change some test cases, like `test_messages_whenOffsetInFileAtEndOfNewestMessageIsBeyondEndOfFile_throwsFileCorrupted`. That case won't get error when `offsetInFileAtEndOfNewestMessage` is beyond EOF.

Sources/CacheAdvance/CacheReader.swift
jianjunwoo: I comment about this at line: https://github.com/dfed/CacheAdvance/pull/72/files#r1013767862

Sources/CacheAdvance/CacheAdvance.swift
dfed: You are correct. The max bytes number is the furthest possible offset but messages within the file might (and in fact likely! end before that point.

Sources/CacheAdvance/CacheReader.swift
dfed: Got it. Because these guards are preventing the seeking behavior which we are no longer doing we don't need them.  🤔. This makes sense. Will dig in more next week!

Sources/CacheAdvance/CacheReader.swift
dfed, line 57: FYI we support all the way back to Xcode 11 (we do this to run tests on older iOS/macOS/etc versions), so we can't use fancy Xcode 14 Swift syntax just yet!

Sources/CacheAdvance/CacheReader.swift
jianjunwoo: Wish you have a nice weekend! and continue the discussion next week : )

bachand:
I will dig into this next week as well. Thanks @jianjunwoo !

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -172,10 +170,28 @@ public final class CacheAdvance<T: Codable> {
         try header.checkFile()
 
         var messages = [T]()
-        while let encodedMessage = try reader.nextEncodedMessage() {
-            messages.append(try decoder.decode(T.self, from: encodedMessage))
-        }
+        // There is only one range: | `offsetInFileOfOldestMessage` -> `offsetInFileAtEndOfNewestMessage`|
+        if reader.offsetInFileOfOldestMessage < reader.offsetInFileAtEndOfNewestMessage {
```
bachand, line 174: When the cache is empty I assume that `reader.offsetInFileOfOldestMessage == reader.offsetInFileAtEndOfNewestMessage`? I wonder if it would be best to explicitly handle the case of these two values being equal, to make this code easier to reason about.

bachand: It would also be nice to validate my assumption that `reader.offsetInFileOfOldestMessage == reader.offsetInFileAtEndOfNewestMessage` when the cache is empty.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -172,10 +170,28 @@ public final class CacheAdvance<T: Codable> {
...
+            for encodedMessage in try reader.encodedMessagesFromOffset(reader.offsetInFileOfOldestMessage,
+                                                                       endOffset: reader.offsetInFileAtEndOfNewestMessage) {
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+        } else {
+            // In this case, the messages could be split to two ranges
+            // | First Range | (GAP: ignore) | Second Range |
```
bachand, line 181: Does there _need_ to be a gap? I think technically the first range could be contiguous with the second range. If it's true that the ranges can be contiguous we should indicate that the gap is optional, for extra clarity.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -23,12 +23,9 @@ final class CacheReader {
 
     /// Creates a new instance of the receiver.
     ///
-    /// - Parameters:
-    ///   - file: The file URL indicating the desired location of the on-disk store. This file should already exist.
-    ///   - maximumBytes: The maximum size of the cache, in bytes.
```
bachand, line 28: 👌

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -44,29 +41,29 @@ final class CacheReader {
...
+            }
+        }
+        if let endOffset = endOffset, offsetInFile != endOffset {
+            throw CacheAdvanceError.fileCorrupted
         }
+        return encodedMessages
+    }
 
+    /// Returns the next encodable message, seeking to the beginning of the next message.
+    func nextEncodedMessage() throws -> Data? {
```
bachand, line 64: Can we make this private now?

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -44,29 +41,29 @@ final class CacheReader {
...
+    /// Returns the encodable messages in a range
+    ///
+    /// - Parameter startOffset: the start offset to start reading
+    /// - Parameter endOffset: the end offset to stop reading, the default value is EOF if the value is nil.
+    func encodedMessagesFromOffset(_ startOffset: UInt64, endOffset: UInt64? = nil) throws -> [Data] {
+        var encodedMessages = [Data]()
+        try reader.seek(to: startOffset)
+        while let data = try nextEncodedMessage() {
+            encodedMessages.append(data)
+            if let endOffset = endOffset, offsetInFile >= endOffset {
```
bachand, line 53: I'm thinking about the case where `offsetInFile == endOffset` to make sure I agree that we want to break in that case.

When we write a message we update the offset in file at the end of the newest message to the offset of the writer after the write.

https://github.com/dfed/CacheAdvance/blob/54e61e242d51e7239bdcf9f4307cd8355b7e86ba/Sources/CacheAdvance/CacheAdvance.swift#L247-L252

I believe that the offset at the end of the newest message is the same as the offset of the start of the next range.

This seems to be confirmed by the [documentation](https://developer.apple.com/documentation/foundation/filehandle/1410936-write) of the write method.

Screen Shot 2022-11-07 at 2 49 04 PM

bachand: @dfed is this the correct documentation to reference? I was surprised to see we are using a deprecated method, though I recall that there was some reason we were doing so.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -75,16 +72,8 @@ final class CacheReader {
             return message
 
         case .emptyRead:
-            guard offsetInFileAtEndOfNewestMessage < startingOffset else {
-                // We started reading before the offset of the end of the newest message, therefore we expected a message to be read. We instead read an empty space, meaning that the file is corrupt.
-                throw CacheAdvanceError.fileCorrupted
-            }
-
-            // We know the next message is at the end of the file header. Let's seek to it.
-            try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)
```
bachand, line 84: I agree that the library is safer against unknown unknown errors without this line of code.

bachand (approved):
Great work @jianjunwoo !

My understanding of this PR is as follows... We aren't currently aware of any type of file corruption that could lead to an infinite loop. At the same time, the way that code is currently written means that an infinite loop remains possible. This PR refactors the library so that an infinite loop is not possible.

It's worth pointing out that so far we haven't (as far as I can recall) acknowledged the two different "modes" of CacheAdvance in the code. The code has been written to transparently handle both situations. There's something elegant with not explicitly acknowledging each case and writing code that will work in either. At the same time, I have found that I spend a lot of time each time I review changes to CacheReader.swift thinking about the different possibilities. I feel like this change makes it easier to consider the edge cases since they are more explicitly captured in code.

I really like that we didn't need to change any tests in this PR. That makes me confident in the change.

Let's give @dfed the time to review as well to make sure he's onboard before we move ahead 👍

Sources/CacheAdvance/CacheReader.swift
jianjunwoo: It could be private, let me do in next commit. 
I kept it here to make the diff more readable.

Sources/CacheAdvance/CacheReader.swift
dfed: these are the `write` methods we're using:
https://github.com/dfed/CacheAdvance/blob/3a1ed6f99c2a870dc09c3a8144a58f915757a569/Sources/CacheAdvance/FileHandleExtensions.swift#L33-L40

I should take a pass to update the `__` methods in this file and use the more modern ones. We are not using deprecated methods when newer ones are available.

Sources/CacheAdvance/CacheAdvance.swift
dfed: I'd need to dig in more, but we have a unit test where we documented that we expect there to be a gap and we do not believe the ranges can be contiguous:
https://github.com/dfed/CacheAdvance/blob/3a1ed6f99c2a870dc09c3a8144a58f915757a569/Tests/CacheAdvanceTests/CacheAdvanceTests.swift#L350-L389

Sources/CacheAdvance/CacheAdvance.swift
dfed: Per this comment block
https://github.com/dfed/CacheAdvance/blob/3a1ed6f99c2a870dc09c3a8144a58f915757a569/Tests/CacheAdvanceTests/CacheAdvanceTests.swift#L367-L368

when `reader.offsetInFileOfOldestMessage` == `reader.offsetInFileAtEndOfNewestMessage`, the file is empty. I like the idea of explicitly handling this case.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -172,10 +170,28 @@ public final class CacheAdvance<T: Codable> {
...
 
+            // This is second range: | `offsetInFileOfOldestMessage` -> EOF |
+            for encodedMessage in try reader.encodedMessagesFromOffset(reader.offsetInFileOfOldestMessage) {
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+
+            // This is first range: | `expectedEndOfHeaderInFile` -> `offsetInFileAtEndOfNewestMessage`|
```
dfed, line 188: nit!
```suggestion
            // This is first range: | `expectedEndOfHeaderInFile` -> `offsetInFileAtEndOfNewestMessage` |
```

Sources/CacheAdvance/CacheAdvance.swift
dfed, line 174: This is nitty, but, let's put the comment within the `if` statement:
```suggestion
        if reader.offsetInFileOfOldestMessage < reader.offsetInFileAtEndOfNewestMessage {
            // There is only one range: | `offsetInFileOfOldestMessage` -> `offsetInFileAtEndOfNewestMessage` |
```

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -172,10 +170,28 @@ public final class CacheAdvance<T: Codable> {
...
+            for encodedMessage in try reader.encodedMessagesFromOffset(
+                FileHeader.expectedEndOfHeaderInFile,
+                endOffset: reader.offsetInFileAtEndOfNewestMessage) {
```
dfed, line 191: let's break this out into a few lines with a named variable:
```suggestion
            let newerMessages = try reader.encodedMessagesFromOffset(
                FileHeader.expectedEndOfHeaderInFile,
                endOffset: reader.offsetInFileAtEndOfNewestMessage)
            for encodedMessage in newerMessages {
```

Sources/CacheAdvance/CacheAdvance.swift
dfed, line 184: Let's break this out into a few lines with a named variable:
```suggestion
            let olderMessages = try reader.encodedMessagesFromOffset(reader.offsetInFileOfOldestMessage)
            for encodedMessage in olderMessages {
```

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -172,10 +170,28 @@ public final class CacheAdvance<T: Codable> {
         try header.checkFile()
 
         var messages = [T]()
-        while let encodedMessage = try reader.nextEncodedMessage() {
-            messages.append(try decoder.decode(T.self, from: encodedMessage))
-        }
+        // There is only one range: | `offsetInFileOfOldestMessage` -> `offsetInFileAtEndOfNewestMessage`|
+        if reader.offsetInFileOfOldestMessage < reader.offsetInFileAtEndOfNewestMessage {
+            for encodedMessage in try reader.encodedMessagesFromOffset(reader.offsetInFileOfOldestMessage,
+                                                                       endOffset: reader.offsetInFileAtEndOfNewestMessage) {
```
dfed, line 176: Let's break this out into a few lines with a named variable:

```suggestion
            let encodedMessages = try reader.encodedMessagesFromOffset(
                reader.offsetInFileOfOldestMessage,
                endOffset: reader.offsetInFileAtEndOfNewestMessage)
            for encodedMessage in encodedMessages {
```

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -44,29 +41,29 @@ final class CacheReader {
...
-    /// Returns the next encodable message, seeking to the beginning of the next message.
-    func nextEncodedMessage() throws -> Data? {
-        let startingOffset = offsetInFile
-
-        guard startingOffset != offsetInFileAtEndOfNewestMessage else {
-            // We're at the last message.
-            return nil
+    /// Returns the encodable messages in a range
+    ///
+    /// - Parameter startOffset: the start offset to start reading
```
dfed, line 46: nitty suggestion:
```suggestion
    /// - Parameter startOffset: the offset from which to start reading
```

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -44,29 +41,29 @@ final class CacheReader {
...
+    /// - Parameter endOffset: the end offset to stop reading, the default value is EOF if the value is nil.
```
dfed, line 47: another nitty suggestion:
```suggestion
    /// - Parameter endOffset: the offset at which to stop reading. If `nil`, the end offset will be the EOF
```

Sources/CacheAdvance/CacheReader.swift
dfed, line 53: When `offsetInFile > endOffset`, wouldn't that mean that the file is corrupted?

EDIT: yup! you covered this case a few lines down 😄

Sources/CacheAdvance/CacheReader.swift
dfed, line 64: I really like that this method is no longer very smart. Our code is more declarative and less imperative now, and the new approach should be simpler to maintain 😄

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -75,16 +72,8 @@ final class CacheReader {
...
-                // We started reading before the offset of the end of the newest message, therefore we expected a message to be read. We instead read an empty space, meaning that the file is corrupt.
-                throw CacheAdvanceError.fileCorrupted
-            }
-
-            // We know the next message is at the end of the file header. Let's seek to it.
-            try reader.seek(to: FileHeader.expectedEndOfHeaderInFile)
-
-            // We know there's a message to read now that we're at the start of the file.
-            return try nextEncodedMessage()
+            // We hit an empty read when reach EOF, so return nil to stop.
```
dfed, line 75: Thoughts on the following comment?
```suggestion
            // An empty read means we hit the EOF. It is the responsibility of the calling code to validate this assumption.
```

dfed (approved):
I left a ton of little nit suggestions, but overall I really dig this approach. I appreciate your spending the time to make this library easier to maintain!

I have also validated that the lines not covered by tests in `CacheReader.swift` are already not covered on `main`.

Sources/CacheAdvance/CacheReader.swift
dfed: Took a pass at updating the methods we use here: https://github.com/dfed/CacheAdvance/pull/74

Sources/CacheAdvance/CacheAdvance.swift
jianjunwoo: like this idea too. handle `==` as empty will make it more clear

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -114,6 +83,26 @@ final class CacheReader {
 
     // MARK: Private
 
+    /// Returns the next encodable message, seeking to the beginning of the next message.
+    private func nextEncodedMessage() throws -> Data? {
```
jianjunwoo, line 87: Mark this func as private since it is not used by other files any more.

Sources/CacheAdvance/CacheAdvance.swift
jianjunwoo: The gap is generated when the new message offset override the old message. Usually the message length of new is not the same as the old message to be rewrote. Then there probably is a gap.

Sources/CacheAdvance/CacheReader.swift
```diff
@@ -50,8 +50,12 @@ final class CacheReader {
...
-                break
+            if let endOffset = endOffset {
+                if offsetInFile == endOffset {
+                    break
+                } else if offsetInFile > endOffset {
+                    throw CacheAdvanceError.fileCorrupted
+                }
             }
         }
         if let endOffset = endOffset, offsetInFile != endOffset {
```
dfed, line 61: do we need this `if` anymore given the `if` within the `while` loop?

jianjunwoo: This make no difference compared to the code before this. However, this looks more readable to handle `==` and `>` separately. @dfed 

```
                else if offsetInFile > endOffset {
                }
```

jianjunwoo: Oooh, I misunderstood your question. The `if` within the `while` loop is need to check `offsetInFile == endOffset `  to stop at the end.

Sources/CacheAdvance/CacheAdvance.swift
bachand: `test_append_inAnOverwritingCacheEvictsMinimumPossibleNumberOfMessages()` proves that there must be a gap. I'm resolving this comment thread.

Sources/CacheAdvance/CacheAdvance.swift
```diff
@@ -172,10 +170,35 @@ public final class CacheAdvance<T: Codable> {
...
+            // There is only one range: | `offsetInFileOfOldestMessage` -> `offsetInFileAtEndOfNewestMessage` |
+            let encodedMessages = try reader.encodedMessagesFromOffset(
+                reader.offsetInFileOfOldestMessage,
+                endOffset: reader.offsetInFileAtEndOfNewestMessage)
+            for encodedMessage in encodedMessages {
+                messages.append(try decoder.decode(T.self, from: encodedMessage))
+            }
+        } else if reader.offsetInFileOfOldestMessage == reader.offsetInFileAtEndOfNewestMessage {
+            // This is an empty cache.
+            return []
```
bachand, line 183: 👍

Sources/CacheAdvance/CacheReader.swift
bachand: > Ahh. This code is catching the `offsetInFile < endOffset` case. Maybe we could make that explicit here? Far from necessary, but it could make the intent more clear if I'm reading this right.

I agree it would be nice to make this more clear.

bachand (approved):
Nice!

jianjunwoo:
Looks like we've all set about the details, or do we still have any other comments that I missed to discuss?

Sources/CacheAdvance/CacheReader.swift
dfed: I'll take this in a follow-up PR

dfed:
That's it! Thanks again @jianjunwoo for tackling this + making future maintenance of this lib easier! Your contributions have immensely improved this library in recent weeks :pray:

PR #73, opened 2022-11-05
Make ObjectiveCError type Sendable without warnings

dfed:
`Error` inherits from `Sendable`, but `NSException` does not conform to `Sendable` (probably due to the `raise()` method?). In order to remove a warning that `ObjectiveCError` does not actually conform to `Sendable` on Xcode 14+, I made `ObjectiveCError` include the exception name and the reason.

Since `ObjectiveCError` is not `public`, this is not a breaking change.

### DIFF
--- a/Sources/CacheAdvance/CacheReader.swift
+++ b/Sources/CacheAdvance/CacheReader.swift
@@ -59,6 +59,7 @@ final class CacheReader {
             }
         }
         if let endOffset = endOffset, offsetInFile < endOffset {
+            // If we finished reading messages but our offset in the file is less than our expected ending offset, this file is corrupted.
             throw CacheAdvanceError.fileCorrupted
         }
         return encodedMessages
--- a/Sources/CacheAdvance/CacheReader.swift
+++ b/Sources/CacheAdvance/CacheReader.swift
@@ -58,7 +58,7 @@ final class CacheReader {
                 }
             }
         }
-        if let endOffset = endOffset, offsetInFile != endOffset {
+        if let endOffset = endOffset, offsetInFile < endOffset {
             throw CacheAdvanceError.fileCorrupted
         }
         return encodedMessages
--- a/Sources/CacheAdvance/CacheReader.swift
+++ b/Sources/CacheAdvance/CacheReader.swift
@@ -54,12 +54,13 @@ final class CacheReader {
                 if offsetInFile == endOffset {
                     break
                 } else if offsetInFile > endOffset {
+                    // The messages on disk are out of sync with our header data.
                     throw CacheAdvanceError.fileCorrupted
                 }
             }
         }
-        if let endOffset = endOffset, offsetInFile < endOffset {
-            // If we finished reading messages but our offset in the file is less than our expected ending offset, this file is corrupted.
+        if let endOffset = endOffset, offsetInFile != endOffset {
+            // If we finished reading messages but our offset in the file is less (or greater) than our expected ending offset, our header data is incorrect.
             throw CacheAdvanceError.fileCorrupted
         }
         return encodedMessages
--- a/Sources/CacheAdvance/ObjectiveC.swift
+++ b/Sources/CacheAdvance/ObjectiveC.swift
@@ -34,7 +34,9 @@ final class ObjectiveC {
         SwiftTryCatch.try({
             result = .success(work())
         }, catch: { exception in
-            result = .failure(ObjectiveCError(exception: exception))
+            result = .failure(ObjectiveCError(
+                exceptionName: exception.name,
+                reason: exception.reason))
         })
 
         switch result {
@@ -51,5 +53,6 @@ struct ObjectiveCTryFailure: Error {}
 
 /// A `throw`able NSException.
 struct ObjectiveCError: Error {
-    let exception: NSException
+    let exceptionName: NSExceptionName
+    let reason: String?
 }
--- a/CacheAdvance.podspec
+++ b/CacheAdvance.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '1.2.4'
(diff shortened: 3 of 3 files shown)


Sources/CacheAdvance/CacheReader.swift
```diff
@@ -58,7 +58,7 @@ final class CacheReader {
                 }
             }
         }
-        if let endOffset = endOffset, offsetInFile != endOffset {
+        if let endOffset = endOffset, offsetInFile < endOffset {
```
dfed, line 61: This change is unrelated to this PR. It addresses [this comment on #72](https://github.com/dfed/CacheAdvance/pull/72/files#r1018595941).

dfed:
@bachand @fdiaz once this PR is approved and merged I'll release the next version bump on Cocoapods.

Sources/CacheAdvance/CacheReader.swift
bachand: To make sure I follow... we'd end up throwing an error from this section of the code if the end offset was past EOF?

bachand: Do you think we're losing any safety here by changing to `<`? I was imagining we'd add a comment to state our intention but leave the check the way it was.

bachand: I can't imagine how we could end up with the `>` here but if we did it seems like that would mean we also have a corrupted file.

dfed: We check for the `>` case a few lines above:
https://github.com/dfed/CacheAdvance/blob/61f099f00e9f01f3bffc2cbab599ecaf36ba51b6/Sources/CacheAdvance/CacheReader.swift#L56-L58

We aren't losing safety today by making this change, and with our current unit tests I don't believe we are losing future safety. Though if someone were to change the unit tests and modify this code we could end up with a future issue, but that doesn't seem terribly likely.

That said, I should absolutely add a comment.

dfed, line 61:
```suggestion
        if let endOffset = endOffset, offsetInFile < endOffset {
            // If we finished reading messages but our offset in the file is less than our expected ending offset, this file is corrupted.
```

Sources/CacheAdvance/ObjectiveC.swift
```diff
@@ -51,5 +53,6 @@ struct ObjectiveCTryFailure: Error {}
 
 /// A `throw`able NSException.
 struct ObjectiveCError: Error {
-    let exception: NSException
+    let exceptionName: NSExceptionName
+    let reason: String?
```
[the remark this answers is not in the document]
dfed: `ObjectiveCError` is considered `Sendable` because `Error` inherits from `Sendable`, and `ObjectiveCError` inherits from `Error`.

`Sendable` requires every member of the conforming type to also be `Sendable` or we get a warning. This change eliminates the warning by ensuring all our members are `Sendable`-conforming types.

We could have alternatively marked this type as `@unchecked Sendable`, but it seemed simpler & better to just make this type truly conform to `Sendable`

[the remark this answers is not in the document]
bachand: Thanks for explaining!

Sources/CacheAdvance/CacheReader.swift
bachand: @dfed the concern I had was that if there was a bug in `nextEncodedMessage()` where we could end up in a situation (in theory) where we got to L61 and `offsetInFile > endOffset`. This could occur if the file were corrupted AND `nextEncodedMessage()` returned `nil` even though it advanced the reader.

Where I personally netted out was that checking `offsetInFile != endOffset` wasn't wrong but resulted in it being difficult for the reader to follow why the check existed. And I felt like that difficulty was resolved by the comment.

I'm open to the idea that this line of thinking is being too cautious though. I wanted to take a moment to voice my thought though I'm also supportive of merging this PR as-is.

bachand: Since @jianjunwoo has been doing great thinking here and authored this code I'll also tag him here for visibility 👍

bachand (approved):

Sources/CacheAdvance/CacheReader.swift
dfed: > the concern I had was that if there was a bug in `nextEncodedMessage()` where we could end up in a situation (in theory) where we got to L61 and `offsetInFile > endOffset`. This could occur if the file were corrupted AND `nextEncodedMessage()` returned `nil` even though it advanced the reader.

This should not be possible, since the only way we return `nil` from `nextEncodedMessage()` is if we get an empty read (which would not advance the reader). That said, that's true today and may not be true forever.

I'll update this code + comment to be a bit more clear, then merge.

jianjunwoo: > "That said, that's true today and may not be true forever."
This is true : D

Overall, it looks fine in either way for me. There are unit tests to cover these cases I think. We've got noticed when it is not true.  

On the other side, when the end offset is larger than the EOF, it could be a warning instead of an error to identify the incorrect endOffset. In this case, the messages are still valid. Currently, we treat this case in a strict way.

dfed: > On the other side, when the end offset is larger than the EOF, it could be a warning instead of an error to identify the incorrect endOffset. In this case, the messages are still valid. Currently, we treat this case in a strict way.

You're right! Prior to #66 we wouldn't have caught this case either. Though because of [the `==` check we have](https://github.com/dfed/CacheAdvance/blob/61f099f00e9f01f3bffc2cbab599ecaf36ba51b6/Sources/CacheAdvance/CacheReader.swift#L54), the only way to hit the `offsetInFile > endOffset` case is for something to have gone _horribly_ wrong (like, more than just an untimely crash), so I think it's valid that we consider files in this state corrupted.

PR #74, opened 2022-11-08
Use most up-to-date FileHandle APIs

dfed:
When iOS 13 / tvOS 13 / watchOS 6 / macOS 10.15 were released, Apple deprecated the [readData(ofLength:)](https://developer.apple.com/documentation/foundation/filehandle/1413916-readdata) and the [write(_:)](https://developer.apple.com/documentation/foundation/filehandle/1410936-write) methods, but _did not replace them with any new documented methods_. The only available, non-deprecated replacements for these methods were `__readDataUp(toLength:)` and `__write(_:, error:)`, but neither were documented.

Then in iOS 13.4 / tvOS 13.4 / watchOS 6.2 / macOS 10.15.4 Apple fixed their mistake, introducing the APIs [read(upToCount:)](https://developer.apple.com/documentation/foundation/filehandle/3516317-read) and [write(contentsOf:)](https://developer.apple.com/documentation/foundation/filehandle/3516320-write). This PR adopts these newer methods.

It is expected that this PR will lower our code coverage, since we do not have CI running on iOS 13.0<=13.4 / tvOS 13.0<=13.4 / watchOS 6<=6.2 / macOS 10.15.0<=10.15.4

### DIFF
--- a/Sources/CacheAdvance/FileHandleExtensions.swift
+++ b/Sources/CacheAdvance/FileHandleExtensions.swift
@@ -23,7 +23,13 @@ extension FileHandle {
 
     /// A method to read data from a file handle that is safe to call in Swift from any operation system version.
     func readDataUp(toLength length: Int) throws -> Data {
-        if #available(iOS 13.0, tvOS 13.0, watchOS 6.0, macOS 10.15, *) {
+        if #available(iOS 13.4, tvOS 13.4, watchOS 6.2, macOS 10.15.4, *) {
+            if let data = try read(upToCount: length) {
+                return data
+            } else {
+                return Data()
+            }
+        } else if #available(iOS 13.0, tvOS 13.0, watchOS 6.2, macOS 10.15, *) {
             return try __readDataUp(toLength: length)
         } else {
             return try ObjectiveC.unsafe { readData(ofLength: length) }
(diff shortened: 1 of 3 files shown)


codecov.yml
```diff
@@ -14,5 +14,6 @@ coverage:
     patch: off
 
 ignore:
+  - "Sources/CacheAdvance/FileHandleExtensions.swift"
```
dfed, line 17: Ignoring this file because due to limitations of Github Actions we can't test most of the branches in this file.

bachand: Can we add a comment to this effect?

Sources/CacheAdvance/FileHandleExtensions.swift
```diff
@@ -23,7 +23,13 @@ extension FileHandle {
 
     /// A method to read data from a file handle that is safe to call in Swift from any operation system version.
     func readDataUp(toLength length: Int) throws -> Data {
-        if #available(iOS 13.0, tvOS 13.0, watchOS 6.0, macOS 10.15, *) {
+        if #available(iOS 13.4, tvOS 13.4, watchOS 6.2, macOS 10.15.4, *) {
+            if let data = try read(upToCount: length) {
+                return data
+            } else {
+                return Data()
```
bachand, line 30: It's interesting that the method both throws and returns a nullable result. I wonder how you could end up with `nil`? In any case, do you think we should throw our own error if the data is `nil`?

bachand (approved):

Sources/CacheAdvance/FileHandleExtensions.swift
dfed: Oh we end up with `nil` when the data is empty. I tried throwing here instead of returning `Data()` and I broke a _ton_ of tests. Returning `Data()` here preserves existing behavior.

codecov.yml
dfed: done!

bachand (approved):

PR #81, opened 2024-01-11
Add support for macCatalyst and visionOS. Drop earlier version support.

dfed:
We were supporting a very old list of Swift, Xcode, and Apple platform versions. This PR brings us into the more modern age, but still keeps a healthy list of supported platforms.


dfed:
Ah shucks. Adding Linux supports requires re-implementing `NSByteOrder`, which is pretty far out of scope. We'll get that on a future pass.

README.md
```diff
@@ -79,12 +79,12 @@ The first 64bytes of a CacheAdvance is reserved for storing metadata about the f
 
 ## Requirements
 
-* Xcode 11.2 or later.
-* iOS 12 or later.
-* tvOS 12 or later.
-* watchOS 5 or later.
-* macOS 10.14 or later.
-* Swift 5.0 or later.
+* Xcode 12.4 or later.
```
dfed, line 82: I'm somewhat tempted to go straight to Swift 5.9 and ignore older Xcode versions, but continuing to support older Xcode versions enables a test suite across more Apple platform versions, and currently there are zero code changes required to support Swift/Xcode versions above Swift 5.5.

Package@swift-5.7.swift
```diff
@@ -0,0 +1,48 @@
+// swift-tools-version:5.7
```
dfed, line 1: these files are just clones of the `Package.swift` unless otherwise specified.

Package@swift-5.9.swift
```diff
@@ -0,0 +1,49 @@
...
+
+let package = Package(
+    name: "CacheAdvance",
+    platforms: [
+        .iOS(.v13),
+        .tvOS(.v13),
+        .watchOS(.v6),
+        .macOS(.v10_15),
+        .macCatalyst(.v13),
+        .visionOS(.v1)
```
dfed, line 14: `visionOS` is new to the Swift 5.9 Package.swift file. We'll add CI for it once Xcode 15.2 comes out.

PR #83, opened 2024-01-31
Run CI on visionOS

dfed:


.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   pod-lint:
     name: Pod Lint
-    runs-on: macOS-13
+    runs-on: macOS-14
     steps:
       - uses: ruby/setup-ruby@v1
```
dfed, line 14: this step is failing right now due to https://github.com/ruby/setup-ruby/issues/568 which will hopefully be resolved soon

.github/workflows/ci.yml
```diff
@@ -130,7 +131,7 @@ jobs:
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_15.0.1.app/Contents/Developer
```
dfed, line 133: while 15.0.1 is supported by the macOS 14 machines, I've seen a ton of spurious failures on those machines so I'm bumping to 15.2 for better support

CacheAdvance.podspec
```diff
@@ -1,15 +1,15 @@
 Pod::Spec.new do |s|
   s.name     = 'CacheAdvance'
-  s.version  = '2.0.0'
+  s.version  = '2.1.0'
   s.license  = 'Apache License, Version 2.0'
   s.summary  = 'A performant cache for logging systems. CacheAdvance persists log events 30x faster than SQLite.'
   s.homepage = 'https://github.com/dfed/CacheAdvance'
   s.authors  = 'Dan Federman'
   s.source   = { :git => 'https://github.com/dfed/CacheAdvance.git', :tag => s.version }
-  s.swift_version = '5.6'
```
dfed, line 9: We compile on Swift 5.6, 5.7, 5.8, and 5.9. To validate that we work on visionOS, we have to lint on Swift 5.9, which we can't do if our podspec specifies an earlier source.

PR #106, opened 2025-09-10
test: Improve test failure messages for isWritable

philipphofmann:
Slightly improve test failure messages for CacheAdvanceTests when the expect fails when checking isWritable, so it's easier to know why a test failed in CI.


dfed (approved):
I'm not convinced this is necessary since the test name is quite clear and `#expect` does a great job of printing the expectation when it fails, but I don't see harm here.

At a minimum: if we're adding comments, I'd prefer the comments to explain the _why_ rather than the _what_, since per the above `#expect` already prints what went wrong on failure. I've left suggestions to that effect.

